PR #10, opened 2018-10-10
Feature/46444 - Add SCSS support via webpack

jszczech:
Resolves #1

PR #11, opened 2018-10-11
Feature/46499 Setup Prettier 

### LINKED ISSUE #9: Setup Prettier
https://github.com/prettier/prettier

jszczech:
Resolves #9


package.json
```diff
@@ -4,7 +4,9 @@
   "version": "0.1.2",
   "scripts": {
     "storybook": "start-storybook -p 6006 -s ./public",
-    "build-storybook": "build-storybook"
+    "build-storybook": "build-storybook",
+    "prettier": "prettier --list-different \"**/*.vue\"",
+    "prettier-fix": "prettier --write \"**/*.vue\""
```
Igloczek: My goal was to connect Prettier with Webpack, to have this as an integrated part of the workflow, to don't even have to think about it.

.prettierrc
```diff
@@ -0,0 +1,9 @@
+{
+  "arrowParens": "avoid",
+  "bracketSpacing": false,
+  "jsxBracketSameLine": false,
+  "semi": false,
+  "singleQuote": true,
+  "tabWidth": 4,
```
Igloczek: We are using 2 spaces in this project

.prettierrc
```diff
@@ -0,0 +1,9 @@
+{
+  "arrowParens": "avoid",
+  "bracketSpacing": false,
+  "jsxBracketSameLine": false,
```
Igloczek: We don't have any JSX in there.

package.json
Igloczek: Check this out - https://medium.com/@doppelmutzi/eslint-prettier-vue-workflow-46a3cf54332f

Will be way better to base prettier on eslint rules, than crate own from scratch.

.storybook/webpack.config.js
```diff
@@ -0,0 +1,11 @@
+module.exports = (storybookBaseConfig, configType, defaultConfig) => {
```
Igloczek: I don't see any reason to have changes from #10 right there.

.eslintrc.js
```diff
@@ -0,0 +1,15 @@
+module.exports = {
+  extends: [
+    "plugin:vue/recommended"
```
Igloczek: We need waaaaaay more stricter rules than this :D

PR #19, opened 2018-11-05
Feature/46956 - Migrate typography

jszczech:


components/01-globals/blocquote/Blockquote.vue
```diff
@@ -0,0 +1,48 @@
+<template>
+  <blockquote>
+    {{ blockquote }}
+    <cite
```
Igloczek: `cite` is not part of `blockquote`

components/01-globals/blocquote/Blockquote.vue
```diff
@@ -0,0 +1,48 @@
+<template>
+  <blockquote>
+    {{ blockquote }}
```
Igloczek: Content should be passed via slot, we don't want to limit users in any way, nor use `v-html`

components/01-globals/blocquote/Blockquote.vue
```diff
@@ -0,0 +1,48 @@
...
+    <cite
+      v-if="cite">
+      <a
```
Igloczek: Links are definitely not a part of the quote element

components/01-globals/grid/GridCollumn.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div :class="['grid__column', width ? `grid__column--${width}` : '']">
+  <div :class="['grid__column', width && `grid__column--${width}`]">
```
Igloczek: This is weird and probably doesn't work anymore, b/c `&&` return Boolean, while we need String.

components/01-globals/preformatted/Preformatted.vue
```diff
@@ -0,0 +1,36 @@
+<template>
+  <component
+    :is="tag ? tag : 'pre'"
+    :class="tag && 'preformatted'">
```
Igloczek: We need to tweak prettier to move closing `>` to the new line.

components/01-globals/preformatted/Preformatted.vue
```diff
@@ -0,0 +1,36 @@
...
+    <slot/>
+  </component>
+</template>
+
+<script>
+export default {
+  props: {
+    tag: {
+      type: String,
+      default: null
```
Igloczek:
```suggestion
      default: 'pre'
```

components/01-globals/preformatted/Preformatted.vue
```diff
@@ -0,0 +1,36 @@
+<template>
+  <component
+    :is="tag ? tag : 'pre'"
```
Igloczek:
```suggestion
    :is="tag"
```

components/01-globals/typography/Heading.vue
```diff
@@ -1,3 +1,12 @@
+<template>
+  <component
+    :is="tag ? tag : 'h' + level"
+    :class="[tag && `heading heading--${levelClassName}-level`,
+             page && 'heading--page']">
```
Igloczek: Dear Prettier, this formatting is 💩 , I'm disappointed :<

components/01-globals/typography/Heading.vue
```diff
@@ -1,3 +1,12 @@
+<template>
+  <component
+    :is="tag ? tag : 'h' + level"
+    :class="[tag && `heading heading--${levelClassName}-level`,
```
Igloczek: Why do you check `tag` if you need `levelClassName`?

components/01-globals/typography/Heading.vue
Igloczek: IMO it should looks like this

```vue
<component
  :is="tag"
  :class="{
    [`heading heading--${levelClassName}-level`]: levelClassName,
    'heading--page': page
  }"
>
```

components/01-globals/typography/Label.vue
```diff
@@ -1,19 +1,18 @@
+<template>
+  <component
+    :is="tag ? tag : 'label'"
+    :class="[tag && 'label', {
```
Igloczek: why `label` is done in array way, but others in the object, while are doing same thing :D?

components/02-elements/button/Button.vue
```diff
@@ -1,34 +1,26 @@
+<template>
+  <component
+    :is="tag ? tag : 'button'"
+    :class="[tag && 'button',
+             customClass && customClass,
+             icon && 'button--icon',
+             secondary && 'button--secondary',
+             link && 'button--link',
+             fluid && 'button--fluid']">
```
Igloczek: It should be in object, not array, will be way cleaner

Igloczek: I'm not sure if that's that option (b/c we don't use JSX), but that's what we need to set to `false` https://prettier.io/docs/en/options.html#jsx-brackets

components/02-elements/button/Button.vue
```diff
@@ -1,34 +1,26 @@
...
+    <slot/>
+    <icon
+      v-if="icon"
+      :icon="icon"
+      :class="'button__icon'"/>
```
Igloczek:
```suggestion
      class="button__icon"
    />
```

components/02-elements/button/Button.vue
```diff
@@ -1,34 +1,26 @@
+<template>
+  <component
+    :is="tag ? tag : 'button'"
+    :class="[tag && 'button',
```
Igloczek: `button` class is not conditional

components/01-globals/typography/Paragraph.vue
```diff
@@ -1,16 +1,17 @@
...
-        class: {
-          paragraph: this.tag
-        }
-      },
-      this.$slots.default
-    )
+  props: {
+    tag: {
+      type: String,
+      default: null
```
Igloczek:
```suggestion
      default: 'p'
```

components/01-globals/typography/Heading.vue
jszczech: B/c when we have a tag, e.g. `h6,` in styles in tag we extend `heading` and `heading--${levelClassName}-level`. It does not make sense to add classes, when we extend this styles inside the tag. If we want to add this classes, I think, we should remove a part of styles. :)

components/01-globals/grid/GridCollumn.vue
jszczech: It works fine this way :)

components/02-elements/button/Button.vue
jszczech: Hmm, why not? If `tag === 'button'`, we don't need add this class.

.eslintrc.js
```diff
@@ -15,6 +15,10 @@ module.exports = {
   },
   rules: {
     "vue/require-default-prop": 1,
-    "vue/require-prop-types": 1
+    "vue/require-prop-types": 1,
+    "vue/html-closing-bracket-newline": [2, {
```
Igloczek: Create an issue in prettier to handle this thing in a correct way

PR #21, opened 2018-11-16
Feature/47806 Migrate badge

jszczech:


components/02-elements/badge/Badge.vue
```diff
@@ -0,0 +1,63 @@
...
+</template>
+
+<script>
+export default {
+  props: {
+    tag: {
+      type: String,
+      default: 'div'
+    },
+    newBadge: {
```
Igloczek: It's kinda weird to have have name of component in prop name, IMO it's perfectly fine to go just with `new` / `discount` or... define a `type` that will be `String` and add a simple validator that checks if value equals something real

components/02-elements/badge/Badge.stories.js
```diff
@@ -0,0 +1,52 @@
...
+      </app>
+    `
+  }))
+  .add('Discount', () => ({
+    components: { App, AlpacaBadge },
+    template: `
+      <app>
+        <alpaca-badge 
+          discountBadge
+         >
```
Igloczek: One space to much in front of `>` 🤓

PR #23, opened 2018-11-20
Feature/47808 Migrate action

jszczech:


components/02-elements/action/Action.vue
```diff
@@ -0,0 +1,104 @@
+<template>
+  <div class="action">
```
Igloczek: Meh, this component in "old" Alpaca obey rules, b/c don't act like an element, but in reality, it's a wrapper that creates a layout for buttons and links.

Let's for now just put it on some "to do" list (idk, maybe some separate directory?) (probably most of the dashboard related stuff will need some love), we will back to it later.

components/02-elements/action/Action.vue
```diff
@@ -0,0 +1,104 @@
...
+    <div
+      class="action__handler"
+    >
+      <alpaca-button
+        v-if="button"
+        :class="[customClass, 'action__button']"
+        @click="click"
```
Igloczek: Not directly related to this component, but general rule

It's better to emit this event outside component, than passing the function inside (I know it's an React thing to do it this way)

So it should look like this

```vue
@click="onClick"
...
methods: {
  onClick(event) {
    this.$emit('click', event)
  }
}
```
and then in the parent component

```vue
<child @click="someEventHandler" />
```

PR #26, opened 2018-11-22
Feature/47934 Migrate banner

jszczech:


components/02-elements/banner/mocks/banner.json
```diff
@@ -1,6 +1,5 @@
 {
   "link": "#",
-  "attributes": "aria-label='descriptive text about banner action and / or image'",
```
Igloczek: We can't just remove it, need an option to define it.

Issue #37, opened 2018-11-22
Migrate Marketing bar


Igloczek:
https://alpaca-components-staging.herokuapp.com/components/detail/marketing-bar.html

PR #48, opened 2018-11-23
Feature/48032

bartdominiak:
internal: Feature/48032
#37

### DIFF
--- a/components/02-elements/marketing-bar/MarketingBar.vue
+++ b/components/02-elements/marketing-bar/MarketingBar.vue
@@ -33,8 +33,7 @@
     },
     data() {
       return {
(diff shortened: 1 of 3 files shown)


components/02-elements/marketing-bar/MarketingBar.vue
```diff
@@ -0,0 +1,121 @@
...
+      }
+    },
+    data() {
+      return {
+        isBarVisible: false
+      }
+    },
+    mounted() {
+      // Display message if it wasn't closed before
+      if (localStorage.getItem(this.dataType) !== 'closed') {
```
kjugi: why just don't set `true/false` value than comparing to string?

PR #53, opened 2018-11-26
#48041: WIP: migrate and refactor price element to price element and …

Januszpl:
WIP: migrating and refactoring price element. 

However as my experience with vue is minimal I'm not sure about about my idea presented it this work in progress PR so waiting for feedback.

1. I think it would be better to split up to single "price" element with different variations like default, special/sale, old/regular as then it's more flexible to build different html structure.
2. Use in price element sth like https://sarahdayan.github.io/dinero.js/ to handle easily different types of showing the price (formatting, currencySymbol ($ at front, zł at the end) or currency (code like USD/PLN), showing FREE (instead of 0 if needed) etc.

3. If required had price-box could/should be refactored still.

### DIFF
--- a/components/02-elements/price/Price.stories.js
+++ b/components/02-elements/price/Price.stories.js
@@ -1,0 +1,81 @@
+import { storiesOf } from '@storybook/vue'
+
+import App from '../../01-globals/app/App.vue'
+import AlpacaPrice from './Price.vue'
+import data from './mocks/price.json';
+
+storiesOf('Elements/Price', module)
+  .add('Default', () => ({
+    data() {
+      const priceData = {
+          ...data.variants.special.price
+      };
+      return {
+        price: priceData,
+        ariaLabel: {
+            "old": "Price reduced from: " + priceData.regularPrice.amount.value + priceData.regularPrice.amount.currency,
+            "special": "Sale starts at: " + priceData.specialPrice.amount.value + priceData.specialPrice.amount.currency
+        }
+      }
+    },
+    components: { App, AlpacaPrice },
+    template: `
+      <app>
+        <alpaca-price type="default">
+          {{ price.regularPrice.amount.currencySymbol }} {{ price.regularPrice.amount.value }}
+        </alpaca-price>
+      </app>
+    `
+  }))
+  .add('Special Price', () => ({
+    components: { App, AlpacaPrice },
+    data() {
+      const priceData = {
+          ...data.variants.special.price
+      };
+      return {
+        price: priceData,
+        ariaLabel: {
+            "special": "Sale starts at: " + priceData.specialPrice.amount.value + priceData.specialPrice.amount.currency
+        }
+      }
+    },
+    template: `
+      <app>
+        <alpaca-price
+          type="special"
+        >
+          <ins :aria-label="ariaLabel.special">
+            {{ price.specialPrice.amount.currencySymbol }} {{ price.specialPrice.amount.value }}
+          </ins>
+        </alpaca-price>
+      </app>
+    `
+  }))
+  .add('Old price', () => ({
+    components: { App, AlpacaPrice },
+    data() {
+      const priceData = {
+          ...data.variants.special.price
+      };
+      return {
+        price: priceData,
+        ariaLabel: {
+            "old": "Price reduced from: " + priceData.regularPrice.amount.value + priceData.regularPrice.amount.currency,
+        }
+      }
+    },
+    template: `
+      <app>
+        <alpaca-price
+          type="old"
+        >
+          <del :aria-label="ariaLabel.old">
+            {{ price.regularPrice.amount.currencySymbol }} {{ price.regularPrice.amount.value }}
+          </del>
+        </alpaca-price>
+      </app>
+    `
+  }));
+
+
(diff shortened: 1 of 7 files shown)


Januszpl:
@Igloczek @jszczech ?

Igloczek:
@Januszpl easy, there is a queue full of tasks that needs some review or feedback.

components/03-modules/catalog-grid-item/CatalogGridItem.vue
```diff
@@ -20,12 +20,31 @@
       </h2>
 
       <div class="catalog-grid-item__price">
```
Igloczek: I'd keep customing the modules and views for another issue.

components/03-modules/price-box/Price-box.vue
```diff
@@ -0,0 +1,25 @@
+<template>
```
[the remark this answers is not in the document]
Januszpl: For element I wouldn't use extra wrapper as not always it can be required and then we increase dom elements in end product or need to overwrite it. Of course we can also skip this component and just add wrapper in places where it will be required.

components/02-elements/price/Price.vue
```diff
@@ -5,48 +5,94 @@
...
+    }
+  },
+  computed: {
+    computedValue() {
+      if(this.prefix) {
+        return this.prefix + this.value
+      }
+      if (this.currency) {
+        return this.value + ' ' + this.currency
       }
```
Januszpl: Why currency at the end? What if we would like to have depending on the currency ($ or zl) at the end or at the front on different store views?

Igloczek: We should get the formatted price from the outside (for example VSF), this is always tricky to decide how to format it.

components/02-elements/price/Price.stories.js
```diff
@@ -0,0 +1,75 @@
...
+            "old": "Price reduced from: " + priceData.regularPrice.amount.value + priceData.regularPrice.amount.currency,
+        }
+      }
+    },
+    template: `
+      <app>
+        <alpaca-price
+          old
+        >
+          <del :aria-label="ariaLabel.old">
```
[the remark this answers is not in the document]
anqaka: yes, it's correct and recommended to use <del> and <ins> for old/special price. Price should have also a label to indicate screen reader that it's the price. I prepare a small codepen to show how it should look like: https://codepen.io/anqa/pen/BMyOrL

anqaka:
For general accessibility purpose we need a label for price value and currency in standard price. When we want to add aria-label as a html prop, the text on aria label will be read instead of the content of it's element, example: 
```
<div aria-label="price">
  $25.90
</div>
```
will be read as just "Price". To fix it we should put the value of the price inside aria-label value. It could be complicated with slots.
In that case I propose to use :before as in the old/special case. 
I put both examples in codepen:
https://codepen.io/anqa/pen/BMyOrL

Igloczek:
@anqaka CSS way seems to be easy, but... translations :) We will need to read some data props in CSS to build pseudoselectors with its content.

components/02-elements/price/Price.vue
```diff
@@ -0,0 +1,80 @@
...
+  props: {
+    ariaLabelSpecial: {
+      type: String,
+      default: 'Special price'
+    },
+    ariaLabelOld: {
+      type: String,
+      default: 'Old price'
+    },
+    specialPrice: {
```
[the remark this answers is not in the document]
szafran89: What about this @jszczech?

components/02-elements/price/Price.vue
```diff
@@ -0,0 +1,87 @@
...
+      if (type === this.oldPrice) {
+        return `${this.ariaLabelOld} ${this.oldPrice}`
+      }
+      return `${this.ariaLabelPrice} ${this.price}`
+    }
+  }
+}
+</script>
+
+<style lang="scss">
```
szafran89: 2 spaces indentation

components/02-elements/price/Price.vue
```diff
@@ -0,0 +1,87 @@
...
+      default: 'Special price:'
+    },
+    ariaLabelOld: {
+      type: String,
+      default: 'Old price:'
+    }
+  },
+  methods: {
+    setAriaLabel(type) {
+      if (type === this.specialPrice) {
```
szafran89: you can use `if` `else if` `else`

components/02-elements/price/Price.vue
Igloczek: We have automatic formatting via eslint enabled, so this is how it's defined in `plugin:vue/recommended`.

components/02-elements/price/Price.vue
Igloczek: It's better to avoid using `else`, when is possible to end execution via `return`

https://softwareengineering.stackexchange.com/questions/206816/clarification-of-avoid-if-else-advice

components/02-elements/price/Price.vue
szafran89: maybe it doesn't work for some reason b/c in some vue components I can see different formatting

components/02-elements/price/Price.vue
szafran89: good point ;)

components/02-elements/price/Price.vue
szafran89: ok, eslisnt doesn't touch styles section

PR #56, opened 2018-11-26
Feature/48021 Migrate input

jszczech:


components/01-globals/label/Label.vue
```diff
@@ -1,11 +1,11 @@
 <template>
   <component
-    :is="tag ? tag : 'label'"
-    :class="{
-      label: tag,
+    :is="tag"
+    :class="[customClass, {
```
Igloczek: We don't need to add a new prop to handle adding a class if it's related to the root level element, b/c it will take the normal `class` value here.

components/02-elements/form/input/Input.vue
```diff
@@ -47,11 +54,15 @@ export default {
       type: String,
       required: true
     },
+    tag: {
```
Igloczek: Since we have just two possible options and one of them is default, IMO will be better to go with Boolean prop `textarea` instead this one.

PR #57, opened 2018-11-26
Feature/48062 Migrate Table

Aekal:


components/02-elements/table/Table.vue
```diff
@@ -0,0 +1,300 @@
+<template>
+  <table
+    class="table"
+    :class="className"
```
jszczech: `:class="['table', className]"` :)

components/02-elements/table/Table.vue
```diff
@@ -0,0 +1,300 @@
...
+    className: {
+      type: String,
+      default: null
+    },
+    caption: {
+      type: String,
+      default: null
+    },
+    headCells: {
+      type: Array,
```
jszczech: I think, headCells should be required.

components/02-elements/table/Table.vue
```diff
@@ -0,0 +1,300 @@
...
+    role="table"
+  >
+    <caption
+      v-if="caption"
+      class="table__caption"
+    >
+      {{ caption }}
+    </caption>
+
+    <thead v-if="headCells">
```
jszczech: I think, heading should be always visible. Maybe if `bodyRows` is missing, we should display some placeholder with information? What do you think?

Aekal: I set both thead and tbody to be required, imo we shouldn't render empty table (without content)

PR #58, opened 2018-11-26
Feature/48058 Migrate radio

jszczech:


components/02-elements/form/radio/Radio.stories.js
```diff
@@ -0,0 +1,26 @@
...
+  .add('Default', () => ({
+    components: { App, AlpacaRadio },
+    data() {
+      return {
+        options
+      }
+    },
+    template: `
+      <app>
+        <alpaca-radio
```
Igloczek: There is no way to get any value from these inputs.

Igloczek: There is no way to get any value from these inputs.

components/02-elements/form/radio/Radio.stories.js
```diff
@@ -20,7 +21,9 @@ storiesOf('Elements/Form/Radio', module)
           :legend="options.legend" 
           :options="options.options" 
           :name="options.name"
+          v-model="selected"
         />
+        Selected option: {{ selected }}
```
Igloczek: I don't think we need it, we can preview this value in Storybook, via https://github.com/Sambego/storybook-state or https://github.com/storybooks/storybook/tree/master/addons/knobs

components/02-elements/form/radio/Radio.vue
```diff
@@ -35,7 +38,15 @@
 
 <script>
   export default {
+    model: {
+      prop: 'checked',
+      event: 'change'
+    },
     props: {
+      checked: {
+        type: [String, Object, Number, Boolean],
```
Igloczek: Huh, why do we need that many types?

components/02-elements/form/radio/Radio.stories.js
Igloczek: But can be done as a enhancement later, so let's just create a task for this and do to next issue :)

components/02-elements/form/radio/Radio.vue
jszczech: This component can accept any of these types.

Igloczek: @jszczech 👀

jszczech: This name is misleading, maybe it should be e.g. 'checkedValue' (i changed it). Values can be different than boolean. 

I checked it in some other library, like bootstrap-vue: 

https://github.com/bootstrap-vue/bootstrap-vue/blob/2e5422aa30cd6197846e66016606d9f3e1455f84/src/components/form-radio/form-radio.js#L29

There also added more types than just boolean.

Issue #39, opened 2018-11-22
Migrate Pager


Igloczek:
https://alpaca-components-staging.herokuapp.com/components/detail/pager.html
PR: #61

PR #61, opened 2018-11-28
Feature/48031

anqaka:
internal dev-48031
issue: #39


components/03-modules/pager/Pager.vue
```diff
@@ -0,0 +1,260 @@
...
+    page: {
+      type: Number,
+      required: true
+    },
+    limit: {
+      type: Number,
+      default: 5
+    },
+    totalSize: {
+      type: Number,
```
jszczech: Please, move required props to the top

components/03-modules/pager/Pager.vue
```diff
@@ -0,0 +1,260 @@
...
+          </alpaca-button>
+        </li>
+      </template>
+
+      <li class="pager__item">
+        <alpaca-button
+          icon="arrow-right"
+          custom-class="pager__link pager__link--next"
+          aria-label="Go to next page"
+          :disabled="currentPage > numberOfPages -1"
```
jszczech: Missing space

components/03-modules/pager/Pager.vue
```diff
@@ -0,0 +1,260 @@
...
+      </li>
+
+      <template v-if="showLast">
+        <li class="pager__item">
+          ...
+        </li>
+        <li class="pager__item">
+          <alpaca-button
+            custom-class="pager__link"
+            @click="currentPage = numberOfPages"
```
jszczech: Maybe we should create methods setCurrentPage ? :)

components/03-modules/pager/Pager.vue
```diff
@@ -0,0 +1,260 @@
...
+      },
+      set (value) {
+        this.$emit('update:page', value)
+      }
+    },
+    numberOfPages () {
+      return Math.ceil(this.totalSize / this.limitPerPage)
+    },
+    listOfPageNumbers () {
+      return Array.apply(null, { length: this.numberOfPages + 1 })
```
jszczech: We can do it shorter:
`Array.from(Array(this.numberOfPages), (d, i) => i + 1)`

components/03-modules/pager/Pager.stories.js
```diff
@@ -0,0 +1,33 @@
+import { storiesOf } from '@storybook/vue'
+
+import App from '../../01-globals/app/App.vue'
+import Pager from './Pager.vue'
+
+import pagination from './mocks/pagination.json'
+
+storiesOf('Modules/Pager', module).add('Default', () => ({
```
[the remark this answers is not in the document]
anqaka: I moved it to modules, b/c it uses elements. The main difference between element and module is that element doesn't use any other element and module does. So this pages use button element so I moved it to modules

components/03-modules/pager/Pager.vue
```diff
@@ -0,0 +1,260 @@
...
+        this.$emit('update:page', value)
+      }
+    },
+    numberOfPages () {
+      return Math.ceil(this.totalSize / this.limitPerPage)
+    },
+    listOfPageNumbers () {
+      return Array.from(Array(this.numberOfPages), (_, i) => i + 1)
+        .map((val, index) => index)
+        .slice(1);
```
jszczech: Could you please remove `map` and `slice` as we don't need them anymore? :)

components/03-modules/pager/Pager.vue
jszczech: IMO it's better to create a function that takes number as an argument and than sets this currentPage. I think, it could be more reusable with the future development in this component.

components/03-modules/pager/Pager.vue
anqaka: done

components/03-modules/pager/Pager.vue
anqaka: updated

PR #64, opened 2018-12-05
Feature/48358 - Migrate additional-content

jszczech:


Igloczek:
Nvm, I already checked. IMO we can remove it and replace with just using grid styles.

PR #66, opened 2018-12-06
Feature/48395 Migrate brief-info

jszczech:


components/03-modules/brief-info/BriefInfo.vue
```diff
@@ -0,0 +1,194 @@
...
+      :level="level"
+      class="brief-info__heading"
+    >
+      {{ heading }}
+    </alpaca-heading>
+
+    <ul class="list brief-info__items-container">
+      <li
+        v-for="(item, i) in items"
+        :key="'brief-info-' + i"
```
Igloczek: Do not use index as key.

components/03-modules/brief-info/BriefInfo.stories.js
```diff
@@ -0,0 +1,22 @@
...
+
+import App from '../../01-globals/app/App.vue'
+import AlpacaBriefInfo from './BriefInfo.vue'
+
+import briefInfo from './mocks/brief-info.json'
+
+storiesOf('Modules/Brief Info', module).add('Default', () => ({
+  components: { App, AlpacaBriefInfo },
+  data: () => ({
+    briefInfo
```
Igloczek: It will be great to rename this component, this name is ultra generic.

components/03-modules/brief-info/mocks/brief-info.json
```diff
@@ -0,0 +1,64 @@
+{
+  "heading": {
+    "text": "Buy with us",
+    "class": ""
+  },
+  "items": [
+    {
+     "iconLabel": "Delivery truck",
+     "iconWrapperClass": "brief-info__icon-wrapper--delivery-truck",
```
Igloczek: I don't like the way how it's done, b/c is not customizable without writing own styles.

This config file should be refactored, we will not get at least 1/3 of props from the backend.

components/03-modules/brief-info/BriefInfo.stories.js
jszczech: Hmm... do you have any idea for the new name? :D

Igloczek: 🤷‍♂️

Igloczek: Let's put it on some TODO list, we will back to this later.

PR #68, opened 2018-12-06
Feature/48311 Migrate active-filters

jszczech:


components/03-modules/active-filters/ActiveFilters.stories.js
```diff
@@ -0,0 +1,24 @@
...
+    components: { App, AlpacaActiveFilters },
+    data: () => ({
+      listElements
+    }),
+    template: `
+      <app>
+        <alpaca-active-filters 
+          :items="listElements.filters"
+          :title="listElements.clear.title"
+          :text="listElements.clear.text"
```
Igloczek: It's not obvious what is the purpose of `title`, `text` and `href` props.

components/03-modules/active-filters/ActiveFilters.vue
```diff
@@ -0,0 +1,168 @@
+<template>
+  <div>
+    <ul class="active-filters list">
+      <li
+        v-for="item in items"
+        :key="item.value"
+        class="active-filters__item list__item"
+      >
+        <alpaca-link
+          href="#"
```
Igloczek: Placeholder.

components/03-modules/active-filters/ActiveFilters.vue
```diff
@@ -0,0 +1,160 @@
...
+        <span class="active-filters__label">
+          {{ item.label }} :&nbsp;
+        </span>
+        <span class="active-filters__value">
+          {{ item.value }}
+        </span>
+      </li>
+    </ul>
+
+    <alpaca-link
```
Igloczek: It's probably going to be a button, no link, so there should be an option to define it somehow + listen for events, to actually clear those filters.

components/02-elements/button/Button.vue
```diff
@@ -69,7 +69,7 @@ export default {
 }
 </script>
 
-<style lang="scss">
+<style lang="scss" scoped>
```
Igloczek: We probably can't do this, b/c styles should be easy to overwrite, there is a high chance, that we will ship it as separate files.

Igloczek: It might make sense to not bind these styles to tags, limit it just to classes, will be easier to create custom buttons.

components/03-modules/active-filters/ActiveFilters.vue
```diff
@@ -0,0 +1,171 @@
+<template>
+  <div>
+    <ul class="active-filters list">
+      <li
+        v-for="item in items"
+        :key="item.value"
+        class="active-filters__item list__item"
```
Igloczek: You should use list component to render list.

components/03-modules/active-filters/ActiveFilters.vue
```diff
@@ -0,0 +1,171 @@
...
+      >
+        <alpaca-link
+          :href="item.href"
+          custom-class="active-filters__remove"
+          :aria-label="item.ariaLabel"
+          :title="item.icon.iconTitle"
+        >
+          <alpaca-icon
+            v-if="item.icon"
+            custom-class="active-filters__remove-icon"
```
Igloczek: I forget about this before, but we need to handle removing a single filter too.

components/03-modules/active-filters/mocks/active-filters.json
```diff
@@ -1,39 +1,32 @@
...
+      "iconId": "close",
+      "iconTitle": "Close"
+    }
+  },
+  {
+    "label": "Label3",
+    "value": "Value3",
+    "ariaLabel": "Disable filter",
+    "href": "#",
+    "icon": {
```
szafran89: close icon shouldn't be defined in mock, we can pass it to the component as an option

szafran89:
We don't need contain in method name string `method`. I propose change it to `clearAll` and `ClearSingleElement` or `ClearItem` or `Clear SingleItem`
image

PR #73, opened 2018-12-19
#48768 - Migrate gallery

jszczech:


components/03-modules/gallery/mocks/gallery.json
```diff
@@ -0,0 +1,52 @@
...
+      }
+    },
+    {
+      "imageThumb": {
+        "dataSrc": "/images/gallery/product-img-80-80.jpg",
+        "alt": ""
+      }
+    }
+  ],
+  "variants": [
```
Igloczek: Do not use the same config files as in Fractal, it's not so readable. If you need two separate variants, create more mock files.

components/03-modules/gallery/mocks/gallery.json
```diff
@@ -0,0 +1,52 @@
+{
+  "class": "",
+  "typeHorizontal": "false",
+  "imageMain": {
+    "dataSrc": "/images/gallery/product-img-496-496.jpg",
+    "alt": "Sample alt"
+  },
+  "thumbs": [
+    {
+      "imageThumb": {
```
Igloczek: what's the point of wrapping everything in `imageThumb` object?

components/03-modules/gallery/mocks/gallery.json
```diff
@@ -0,0 +1,52 @@
+{
+  "class": "",
+  "typeHorizontal": "false",
```
Igloczek: It's not a place to keep stuff like this, mocks should be used just to mimic the backend responses.

components/03-modules/gallery/Gallery.vue
```diff
@@ -0,0 +1,204 @@
...
+    <div :class="['gallery__nav', horizontal ? 'gallery__nav--horizontal' : 'gallery__nav--vertical']">
+      <div class="gallery__thumb-arr">
+        <alpaca-icon
+          icon="angle-up"
+          :class="['gallery__icon-arrow', horizontal ? 'gallery__icon-arrow--horizontal' : 'gallery__icon-arrow--vertical']"
+        />
+      </div>
+      <div
+        v-for="(thumb, i) in thumbs"
+        :key="i"
```
Igloczek: It's not a good practice to use an item index as keys, because if you remove/move one element, all other items will get a new key (b/c index was changed) and will be re-rendered, even if nothing was changed.

https://vuejs.org/v2/guide/list.html#key
https://forum.vuejs.org/t/v-for-with-simple-arrays-what-key-to-use/13692/11

components/03-modules/gallery/Gallery.vue
```diff
@@ -0,0 +1,204 @@
...
+      AlpacaBadge,
+      AlpacaImage,
+      AlpacaIcon
+    },
+    props: {
+      imageMain: {
+        type: Object,
+        required: true
+      },
+      thumbs: {
```
Igloczek: IMO we should get just one array of images, but in two sizes. Clicking on the thumbnail should change the main image.

PR #77, opened 2019-01-10
Feature/49084 Migrate quantity-update

jszczech:


components/02-elements/form/file-upload/FileUpload.vue
```diff
@@ -51,7 +51,7 @@
 </template>
 
 <script>
-import AlpacaButton from "../../../02-elements/button/Button.vue";
+import AlpacaButton from "../../../02-elements/button/Button.vue"
```
szafran89: There are sometimes extensions `.vue` in import line. If we don't need it please remove.

PR #93, opened 2019-02-02
Feature/49812 Migrate information

jszczech:


components/02-elements/information/Information.stories.js
```diff
@@ -0,0 +1,24 @@
...
+  .add('Default', () => ({
+    data: () => ({
+      information
+    }),
+    components: { App, AlpacaInformation },
+    template: `
+      <app>
+        <alpaca-information
+          v-for="info in information"
+          :ke="info.id"
```
szafran89: typo?

components/02-elements/information/Information.vue
```diff
@@ -0,0 +1,49 @@
...
+      text: {
+        type: String,
+        required: true
+      }
+    }
+  }
+</script>
+
+<style lang="scss">
+  $information-margin: 15px !default;
```
szafran89: should be `$information__margin`

szafran89: can we use `$spacer--medium` instead of this one

PR #95, opened 2019-02-05
Feature/49813 Migrate quick-search

jszczech:


components/03-modules/quicksearch/mocks/manufacturers.json
```diff
@@ -0,0 +1,26 @@
...
+    "id": "manufacturer3",
+    "title": "Manufacturer name",
+    "href": "#",
+    "items": "598 item(s)"
+  },
+  {
+    "id": "manufacturer4",
+    "title": "Manufacturer name",
+    "href": "#",
+    "items": "598 item(s)"
```
szafran89: keep only number here, move suffix `item(s)` to template. Apply in all places

PR #96, opened 2019-02-06
Feature/48421 Migrate dashboard

jszczech:


Aekal:
Please refactor these components to our new approach with slots and move to proper folders according to atomic design. Then I'll check if further

PR #99, opened 2019-02-08
Feature/49346 Migrate catalog-grid-item

jszczech:


components/03-modules/catalog-grid-item/CatalogGridItem.vue
```diff
@@ -1,124 +1,237 @@
...
+    :is="tag"
+    class="catalog-grid-item"
+  >
+    <alpaca-badge
+      v-if="badgeText"
+      :type="badgeType"
+    >
+      {{ badgeText }}
+    </alpaca-badge>
+    <alpaca-link
```
szafran89: router link instead of alpaca link?

components/03-modules/catalog-grid-item/CatalogGridItem.vue
```diff
@@ -1,124 +1,237 @@
...
-      >
-    </a>
-
+      />
+    </alpaca-link>
     <section class="catalog-grid-item__details">
       <h2 class="catalog-grid-item__name">
-        <a
-          href="#"
+        <alpaca-link
```
szafran89: router link instead of alpaca link?

PR #102, opened 2019-02-11
Feature/49898 Migrate footer module

jszczech:


components/02-elements/dropdown-list/dropdown-list-link/DropdownListLink.html
```diff
@@ -1,5 +1,5 @@
 <li class="dropdown-list__item">
-  <a class="dropdown-list__label" :href="href">
+  <a :class="['dropdown-list__label', linkClass]" :href="href">
```
szafran89: I think we should use `route-link`

PR #104, opened 2019-02-11
Feature/50106 Fix bugs in catalog component

jszczech:


components/02-elements/price/Price.vue
```diff
@@ -27,15 +27,15 @@
   export default {
     props: {
       price: {
-        type: String,
+        type: [String, Number],
```
szafran89: Imo should be String only

PR #106, opened 2019-02-14
Feature/48804 Migrate dropdown-list

jszczech:


szafran89:
IMO we don't need this arrow http://prntscr.com/mkxk98 and wrong color on hover

components/02-elements/dropdown-list/DropdownList.js
```diff
@@ -0,0 +1,28 @@
+export default {
+  provide: {
+    nested: true
+  },
+  inject: {
+    nested: {
+      default: false
+    }
+  },
```
szafran89: I'm not fan of this solution, can we change it to regular props

Issue #109, opened 2019-02-19
Add webpack stylelint


robwozniak:
https://github.com/webpack-contrib/stylelint-webpack-plugin

PR #110, opened 2019-02-19
#109 - Add stylelint-webpack plugin with base stylelint configuration

robwozniak:


.storybook/webpack.config.js
```diff
@@ -34,5 +36,13 @@ module.exports = (storybookBaseConfig, configType, defaultConfig) => {
     }
   })
 
+  defaultConfig.plugins.push(new StyleLintPlugin({
+    files: [
+      'components/**/*.vue',
+      'assets/styles/**/*.scss'
+    ],
```
szafran89: we have also `scss` styles in components directory. e.g `/components/02-elements/button/Button.scss` We are going to divide all components to separate files so I'm not sure if you cover these files?

szafran89:
Now we have failing checks. Please fix it ;)

robwozniak:
Ok, I thought that we can handle that in the other task :smile:

.storybook/webpack.config.js
robwozniak: Good catch! These files were omitted :+1:

PR #116, opened 2019-02-21
Feature/50384 Update grid component

jszczech:


docs/styles/_grid.scss
```diff
@@ -0,0 +1,23 @@
+.docs-only-styles {
+    .col-large .row__content {
```
szafran89: indentation, and add to stylelint this file

components/01-globals/grid/Grid.html
```diff
@@ -0,0 +1,20 @@
+<div>
+  <alpaca-heading :level="1">{{ title }}</alpaca-heading>
```
szafran89: content to new line?

szafran89: Do we need this heading, I can't see this heading http://prntscr.com/moirpg

szafran89:
Can we remove `grid-column-width` mixin?
Can we remove `assets/styles/_grid.scss` file?
Do we need any adjustment for existing components?

PR #118, opened 2019-02-24
Migrate range

talalus:
internal-dev: 49695


components/03-modules/range/Range.vue
```diff
@@ -0,0 +1,261 @@
...
+      apply() {
+        this.$emit('apply', this.currentValue)
+      }
+    }
+  }
+</script>
+
+<style lang="scss">
+  $range__padding       : $spacer 0 0 0 !default;
+  $range__input-margin  : 0 !default;
```
jszczech: Please, use $reset variable

components/03-modules/range/Range.vue
```diff
@@ -0,0 +1,261 @@
...
+      return {
+        currentValue: this.value
+      }
+    },
+    computed: {
+      minInput: {
+        get: function() {
+          return this.currentValue[0]
+        },
+        set: function(minValue) {
```
jszczech: Please, compare behavior with https://alpaca-components.now.sh/components/detail/range.html
It's a little strange, that we can move the min value to the max and then it becomes the max :D

talalus:
vue-slider value type is Array or number https://nightcatsama.github.io/vue-slider-component/#/api/props but we emit the value as object if this is required

PR #121, opened 2019-03-04
Fix instructions based on package.json

youanden:


youanden:
I see, master branch uses "storybook", but `develop` branch uses `dev`

I would recommend updating master in that case.

PR #123, opened 2019-03-05
feature/50825 - Wishlist module

robwozniak:


components/03-modules/minicart/Minicart.html
```diff
@@ -1,16 +1,17 @@
...
-      :name="`qty_${key}`"
+      :product-image="product.productImage"
+      :product-name="product.productName"
+      :product-price="product.price"
+      :product-id="product.productId"
+      :product-options="product.productOptions"
+      :quantity="product.quantity"
+      :remove-btn="product.removeBtn"
+      :name="`minicart_${key}`"
       href="#"
```
jszczech: Should it be hardcoded in this way here?

components/03-modules/product-item/ProductItem.html
```diff
@@ -0,0 +1,72 @@
...
+        class="product-item__qty-wrapper"
+      >
+        <span class="product-item__qty-text">
+          {{ quantity.text }}
+        </span>
+        <alpaca-quantity-update
+          :input="quantity.input"
+          :minus-button="quantity.minusQtyButton"
+          :plus-button="quantity.plusQtyButton"
+          decrement-button-class="product-item__qty-button"
```
jszczech: IMO will be better to use minus or decrement in all places, not both. The same in case of plus/increment :D

components/03-modules/minicart/Minicart.js
```diff
@@ -2,21 +2,21 @@ import uniqueId from 'lodash.uniqueid'
...
 export default {
   components: {
     AlpacaHeading,
     AlpacaList,
     AlpacaButton,
-    AlpacaCartListItem
+    AlpacaProductItem
   },
   props: {
     taxAmount: {
```
jszczech: Please, keep required props on the top

robwozniak:
Ad. 1. Done
Ad. 2. If line with classes do not exceed 80 chars then indentations are not needed.
Ad. 3. Fixed :slightly_smiling_face:

PR #137, opened 2019-03-22
Feature/51613 - add trap focus in modules Modal and OffCanvasSidebar

niemaco:


components/03-modules/modal/Modal.js
```diff
@@ -69,36 +67,55 @@ export default {
       this.ariaHidden = state ? 'false' : 'true'
       this.visibility = state
 
-      if (state === true) {
-        this.$nextTick(() => this.$refs.modal.focus())
-      } else {
+      if(state) {
         this.$nextTick(() => this.focused.focus())
+      } else {
+        this.$nextTick(() => this.trigger.focus())
```
szafran89: please check if we can get rid of this condition

components/03-modules/modal/Modal.js
```diff
@@ -69,36 +67,55 @@ export default {
       this.ariaHidden = state ? 'false' : 'true'
       this.visibility = state
 
-      if (state === true) {
-        this.$nextTick(() => this.$refs.modal.focus())
-      } else {
+      if(state) {
```
szafran89: missing space after `if`

niemaco: I removed the above code.

components/03-modules/modal/Modal.js
niemaco: I removed unnecessary data "focused" and above statement. I've moved the appropriate focus setting to show and hide methods.

PR #141, opened 2019-03-26
Feature/50030 Fix mobile styles and use alpaca components for header

jszczech:


components/src/03-modules/header-search/HeaderSearch.scss
```diff
@@ -21,6 +21,7 @@ $search-form__placeholder-color : $gray !default;
   border: 1px solid $search-form__border-color;
   border-radius: $search-form__height;
   z-index: 1;
+  position: relative;
```
szafran89: position should be on the top, above `width`
It should look more or less like this, most important are 2 first points
```
Layout Properties (position, float, clear, display)
Box Model Properties (width, height, margin, padding)
Visual Properties (color, background, border, box-shadow)
Typography Properties (font-size, font-family, text-align, text-transform)
Misc Properties (cursor, overflow, z-index)
```

PR #142, opened 2019-03-27
Feature/51810 Missing transitions and wrong styles for header

jszczech:


components/src/01-globals/grid/Grid.stories.js
```diff
@@ -9,28 +9,31 @@ storiesOf('Global/Grid', module)
...
+               ]
+             ]"
+          />
+      </div>
     `
   }))
   .add('Fluid', () => ({
     components: { AlpacaGrid },
     template: `
+      <div class="wrapper">
```
szafran89: What do you think to use storybook class `sb-show-main` instead of adding new `wrapper`.

components/src/01-globals/grid/Grid.stories.js
```diff
@@ -9,28 +9,31 @@ storiesOf('Global/Grid', module)
...
-              'col-xs-12 col-sm-3 col-md-2 col-lg-1',
-              'col-xs-12 col-sm-9 col-md-10 col-lg-11'
-            ],
-            [
-              'col-xs-10 col-sm-6 col-md-8 col-lg-10',
-              'col-xs-2 col-sm-6 col-md-4 col-lg-2'
-            ]
-          ]"
-        />
+       <div class="wrapper">
```
szafran89: there is sth wrong with our grid http://prntscr.com/n3pc75

PR #145, opened 2019-03-29
Feature/51905

Aekal:


components/.storybook/config.js
```diff
@@ -20,6 +20,12 @@ addDecorator(() => ({
   `
 }))
 
+addParameters({
+  options: {
+    name: 'Alpaca components'
```
Aekal: Do we want to stay with `/components` dir and title name or change it to `/ui`?

szafran89: please keep `/component` for now, we'll think it over on next alpaca meeting :)

PR #146, opened 2019-04-02
Feature/52059 - Select

Aekal:
dev-52059


components/src/molecules/select/Select.html
```diff
@@ -1,15 +1,14 @@
...
     :class="['select__label', labelClass]"
   >
-    {{ label }}
-   </a-label>
+    <!-- @slot Label text -->
+    <slot />
+  </a-label>
   <div class="select__field">
     <select
-      :id:="id"
```
szafran89: we need support for `v-model`. Please consider if we should use scoped slots or there is any other options

components/src/molecules/select/Select.stories.js
```diff
@@ -0,0 +1,55 @@
...
+    `,
+    computed: {
+      countryOptions() {
+        return countries.map((country) => {
+          return {
+            value: country.code,
+            text: country.name
+          }
+        })
+      }
```
szafran89: move `countryOptions` before stories to reduce duplication

components/src/molecules/select/Select.stories.js
```diff
@@ -0,0 +1,55 @@
+import {storiesOf} from '@storybook/vue'
```
szafran89: checnge to `{ storiesOf }`

components/src/molecules/select/Select.stories.js
```diff
@@ -0,0 +1,55 @@
...
+  }))
+  .add('Hidden label', () => ({
+    components: { ASelect },
+    template: `
+      <a-select
+        id="field-id"
+        name="field-name"
+        :options="countryOptions"
+        selected="PL"
+        labelClass="label--hidden"
```
szafran89: https://eslint.vuejs.org/rules/attribute-hyphenation.html

components/src/molecules/select/Select.stories.js
```diff
@@ -0,0 +1,55 @@
...
+        name="field-name"
+        :options="countryOptions"
+        selected="PL"
+      >
+        Label text
+      </a-select>
+    `,
+    computed: {
+      countryOptions() {
+        return countries.map((country) => {
```
szafran89: can we change mock instead of parsing this data?

szafran89:
Additionally we need support for default option, you can also add selected option under rendered component inside stories, sth like here https://bootstrap-vue.js.org/docs/components/form-select

components/src/molecules/select/Select.html
```diff
@@ -29,11 +28,11 @@
...
-        :key="index"
-      >
-        {{ validation.text }}
-      </span>
+    <span
+      v-for="(validation, index) in currentValidations"
+      :key="index"
+    >
+      {{ validation.text }}
+    </span>
```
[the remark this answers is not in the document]
Aekal: ok, removed

components/src/molecules/select/Select.html
Aekal: support for v-model added, you can see changing value in actions tab, it now works similar to vsf `BaseSelect`

components/src/molecules/select/Select.stories.js
Aekal: removed, using mock instead

components/src/molecules/select/Select.stories.js
Aekal: done

components/src/molecules/select/Select.stories.js
Aekal: I always forgot about it, that's why we need eslint :D

components/src/molecules/select/Select.stories.js
Aekal: done

components/src/molecules/select/Select.html
Aekal: I'm not sure if it's better to show value in actions tab or directly under select component as we have in `swatch` component, what do you prefer?

Aekal:
Ok, I've finally got it working with `v-model` and `@change` event. Variant with unselected option added

components/src/molecules/select/Select.html
szafran89: actions addon is ok.

components/src/molecules/select/Select.html
```diff
@@ -1,39 +1,32 @@
 <div class="select">
   <a-label
     :for="id"
-    :hidden="hiddenLabel"
-    :inline="inlineLabel"
     :class="['select__label', labelClass]"
   >
-    {{ label }}
-   </a-label>
+    <!-- @slot Label text -->
```
szafran89: Maybe better `Label content`?

components/src/molecules/select/Select.html
```diff
@@ -1,39 +1,32 @@
...
       @change="$emit('input', $event.target.value)"
     >
-      <option v-if="!selected" />
+      <option disabled selected v-if="!value">
+        <!-- @slot Default option text -->
+        <slot name="option"/>
+      </option>
       <option
         v-for="(option, key) in options"
         :key="key"
```
szafran89: use `option.value` and remove key from `v-for`

components/src/molecules/select/Select.stories.js
```diff
@@ -0,0 +1,89 @@
+import { storiesOf } from '@storybook/vue'
+import { action } from '@storybook/addon-actions'
+
+import countries from '../../../mocks/countries.json'
+
+import ASelect from './Select.vue'
+
+storiesOf('Molecules/Select', module)
+  .addParameters({ info: true })
+  .add('Default', () => ({
```
szafran89: Please format the same way as radio/checkbox

Aekal: I'm not a huge fan of this formatting, but ok, done :P

components/src/molecules/select/Select.stories.js
szafran89: Yes, but IMO will not work inside story

components/src/molecules/select/Select.stories.js
```diff
@@ -0,0 +1,98 @@
...
+          id="field-id"
+          name="field-name"
+          :options="countries"
+          v-model="selectedCountry"
+          @input="showCountry"
+        >
+          <template v-slot:label>
+            Label text
+          </template>
+          <template v-slot:option>
```
szafran89: please use shortcut from vue 2.6.0 in all places, then can go to develop

Aekal: Done

PR #151, opened 2019-04-04
Feature/52173 Refactor logo component

jszczech:


components/src/molecules/logo/Logo.stories.js
```diff
@@ -0,0 +1,36 @@
...
+    'With slots',
+    () => ({
+      components: { ALogo },
+      template: `
+        <a-logo to="/">
+          <template #image>
+            <img 
+              src="/images/logo/alpaca.svg" 
+              alt="Logo"
+              :style="{width: '70%'}"
```
szafran89: why you can't use regular form `style="width: 70%;"`?

components/src/molecules/logo/Logo.stories.js
```diff
@@ -0,0 +1,36 @@
+import { addDecorator, storiesOf } from '@storybook/vue'
+import StoryRouter from 'storybook-vue-router'
+
+import ALogo from './Logo.vue'
+
+addDecorator(StoryRouter())
```
szafran89: can you add decorator under storiesOff?

components/src/molecules/logo/Logo.stories.js
jszczech: I can :D

components/src/molecules/logo/Logo.stories.js
jszczech: Did you mean this way? 

`storiesOf('Modules/Side Menu', module)
  .addDecorator(StoryRouter())`

Firstly I tried do it, but than we have a problem with order of imports , and get a errors. We can add it globally in config, and it will work, but I don't know if this is a good way.

szafran89: ok so keep it in this way

PR #155, opened 2019-04-09
feature/51310 - eslint configuration

robwozniak:


Aekal:
This solution with `// @vue/component` comment is nice, we should try to add it also to `.html` and `.stories.js` files to have vue linting everywhere :D

szafran89:
@robwozniak solution is ok, IMO is better to use `// @vue/component` rather than `jsx`. Please try to apply lint to `html` and `stories` files as @Aekal has written one comment before

robwozniak:
Checked, unfortunately there is no equivalent for `@vue/component` comment inside html files. We could move template content directly to the vue files.

Anyways, vue linter was added to stories files with success :wink:

PR #156, opened 2019-04-10
Feature/52206 Refactor dropdown list

jszczech:


components/src/atoms/dropdown-list-link/DropdownListLink.html
```diff
@@ -0,0 +1,12 @@
+<li class="dropdown-list__item">
```
Aekal: I know it looks like an atom, but as far as it's a part of `DropdownList` component I think it should be moved to `/molecules/dropdown-list/link/`, the same for `DropdownListItem` - can be moved to `/molecules/dropdown-list/item/`. I think we should group components in this way, they also have shared styles in main component.
cc @szafran89 what do you think?

components/src/molecules/dropdown-list/DropdownList.html
```diff
@@ -1,19 +1,15 @@
-<div :class="['dropdown-list', {
-  'dropdown-list--inner': inner,
-  'dropdown-list--secondary': secondary,
-  'dropdown-list--dark': dark,
-  'dropdown-list@screen-m dropdown-list--is-open@screen-m': screenM
-  }]"
->
+<div :class="['dropdown-list', { 'dropdown-list--inner': nested }]">
```
Aekal: `dropdown-list--inner` class can be passed in story, I think we can leave here only `class="dropdown-list"`

components/src/molecules/dropdown-list-item/DropdownListItem.js
```diff
@@ -5,22 +5,37 @@ export default {
...
-      required: true
+      required: null
     },
+    /**
+     * Item collapse
+     */
     collapse: {
       type: Boolean,
-      default: true
+      default: null
```
Aekal: It's a boolean so I suggest to change it to `false`

components/src/molecules/dropdown-list-item/DropdownListItem.html
```diff
@@ -5,18 +5,22 @@
     @click="toggle"
```
Aekal: In this file
```
<li :class="['dropdown-list__item', {'dropdown-list__item--collapse': collapseLocal}]">
```
we should change it to 
```
<li class="dropdown-list__item">
```
and add collapse class to info

components/src/molecules/dropdown-list-item/DropdownListItem.html
```diff
@@ -5,18 +5,22 @@
...
+    </slot>
   </button>
   <div
     :id="id"
     :class="['dropdown-list__content', contentClass]"
     :data-content="id"
     :aria-hidden="collapseLocal"
   >
+    <!--@slot Default content-->
     <slot />
```
Aekal: Maybe let's add a name content to this slot to make it more descriptive

components/src/atoms/dropdown-list-link/DropdownListLink.html
jszczech: Initially it was done in this way, but later we assumed that everything should be on one level. 
In alpaca channel was wrote:
> one component - one directory, now we have some sets of components like table, dropdown-list etc.
But maybe we should discuss it again.

components/src/molecules/dropdown-list/DropdownList.html
jszczech: We can pass `dropdown-list--inner` in the story, if we want  `inner styles`, but here it's for nested dropdown. During using it we can't pass this modifirer just in root, because it will be used for all dropdown, we need it only for nested elements.

components/src/molecules/dropdown-list-item/DropdownListItem.html
jszczech: But this variable is dependent on logic inside the component.

components/src/atoms/dropdown-list-link/DropdownListLink.html
Aekal: Ok, let's stay with this convention for now

Aekal:
I'm not sure if should keep component specific data in `DropdownList.stories.js`, I think we can move it to `mocks` inside the `dropdown-list/` for better readibility
@jszczech @szafran89

szafran89:
If those data should comes from backen we have to move it to mock @Aekal @jszczech

components/src/molecules/dropdown-list/DropdownList.html
```diff
@@ -1,19 +1,15 @@
...
   <ul :class="['dropdown-list__list', listClass]">
+    <!-- @slot Default content -->
     <slot v-if="!items" />
+    <!--Default content passed with props-->
     <template v-else v-for="element in items">
```
szafran89: @jszczech we can not pass `v-if` `v-else` together with `v-for`

components/src/molecules/collapse/Collapse.html
```diff
@@ -0,0 +1,24 @@
+<div class="a-collapse">
+  <button
```
Aekal: I think this entire button should be inside scoped slot (with @click event and `collapseLocal` passed through scope).
Then you can remove `titleClass` and move title to prop instead of using slot - for cases where button slot won't be used

components/src/molecules/collapse/Collapse.html
```diff
@@ -0,0 +1,24 @@
...
+    :aria-controls="collapseLocal"
+    @click="toggle"
+    :aria-expanded="!collapseLocal"
+  >
+    <!-- @slot Custom title slots (Named slot) -->
+    <slot name="title">
+      {{ title }}
+      <a-icon
+        icon="angle-down"
+        class="a-collapse__icon"
```
Aekal: Add `title="Angle down"` attribute

components/src/molecules/collapse/Collapse.scss
```diff
@@ -0,0 +1,150 @@
...
+    display: block;
+    width: $collapse__button-width;
+    margin: $reset;
+    padding: $collapse__button-padding;
+    color: $black;
+    text-align: left;
+    text-decoration: none;
+    border: 0;
+    cursor: pointer;
+
```
Aekal: Add base transition and please update icon color to green on hover
Zrzut ekranu 2019-05-15 o 12 45 17

components/src/molecules/collapse/Collapse.spec.js
```diff
@@ -0,0 +1,50 @@
+import { mount } from '@vue/test-utils'
+import ACollapse from './Collapse.vue'
+
+describe('Button', () => {
```
Aekal: `describe('Collapse'`

components/src/molecules/collapse/Collapse.html
```diff
@@ -1,20 +1,27 @@
...
-    :aria-controls="collapseLocal"
-    @click="toggle"
+  <!-- @slot Custom button slots (Scoped slot) -->
+  <slot
+    name="button"
+    :toggle="toggle"
+    :collapseLocal="collapseLocal"
+    :ariaControls="ariaControls"
+  >
+  <a-button
```
Aekal: Indentation

components/src/molecules/collapse/Collapse.html
```diff
@@ -1,20 +1,27 @@
...
+    class="a-collapse__button a-button--blank"
+    :aria-controls="ariaControls"
+    @click.native="toggle"
     :aria-expanded="!collapseLocal"
   >
-    <!-- @slot Custom title slots (Named slot) -->
-    <slot name="title">
       {{ title }}
       <a-icon
+        title="Angle down"
```
Aekal: Btw this title should be in prop for supporting translation, but I think we can skip if for now. Afaik we'll change soon `Icon` component and will be best to have the icon title defined in `Icon` component level.

components/src/molecules/collapse/Collapse.stories.js
```diff
@@ -53,9 +53,15 @@ storiesOf('Molecules/Collapse', module)
         ACollapse
       },
       template: `
-        <a-collapse>
-          <template #title>
-            <span>Title</span>
+        <a-collapse title="Sample content">
```
Aekal: `title="Sample content"` is not rendered when button slot is used, you can remove it here.
But after removing this not required prop `title` we get an error related to `ariaControls`, please check it

PR #161, opened 2019-04-16
#52468 - Remove lodash.uniqueid

Aekal:


components/src/03-modules/gallery/Gallery.html
```diff
@@ -24,8 +24,8 @@
       />
     </div>
     <div
-      v-for="({ key, image }, i) in getThumbsWithKey"
-      :key="key"
+      v-for="(image, i) in images"
+      :key="i"
```
jszczech: The same situation as in `BriefInfo.html`. It was a reason why I used it.

components/src/03-modules/brief-info/BriefInfo.html
```diff
@@ -9,7 +9,7 @@
 
   <ul class="list brief-info__items-container">
     <li
-      v-for="{ key, item } in getItemsWithKey"
+      v-for="(item, key) in items"
```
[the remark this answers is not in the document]
Aekal: Please ignore it for now in this component. It needs to be refactored later, we'll remove mock and use slot with list items instead (so we won't need v-for at all). Let's just remove lodash for now.

components/src/03-modules/gallery/Gallery.html
Aekal: I did a research about `:key` and found these two resources:
- [The Key to the Vue v-for](https://rimdev.io/the-v-for-key/)
> The tl;dr from all this is that it’s easy to just use the index and not run into any problems as long as you’re always appending to the end, or have a simple list. This is a great way to do things (and the default used by Vue) as long as you aren’t doing more complex list manipulation. But the moment that you start having state in the component itself that relies on the prop, or changing the lists in a way that isn’t just tacking things on at the end, use a unique id.

- https://github.com/vuejs/vue/issues/6235#issuecomment-318607572
> If your data don't have any field that could be used as unique keys, there is something not right about it, ask your backend team.

In vuestorefront I see this gallery data from backend:
Zrzut ekranu 2019-04-16 o 23 59 05
Some images have id but some don't. 
In this case we can:
1. Use id from backend as a `key`, counting on that every image will have unique id
2. Use index as a `key`, probably we'll use list in a standard way without complex manipulation here
3. Get file name from `src` and use as a `key` e.g. `wt06-red_main.jpg`. I think every file name is unique in each product, but I'm not 100% about that.

What do you think, which solution will be best for this case? cc @szafran89 
Sorry for that long comment but I'm curious how to handle `key` in general :D

szafran89: Please use id from backend and assume that every image have unique id.

PR #163, opened 2019-04-17
Feature/52520 Refactor breadcrumbs

jszczech:


components/src/04-views/catalog/Catalog.stories.js
```diff
@@ -27,7 +26,24 @@ storiesOf('Views/Catalog', module)
         activeFilterItems,
         checkboxItems,
         toolbar,
-        breadcrumbs
+        breadcrumbs: [
```
Aekal: Let's move it to `components/mocks/breadcrumbs.json` and use globally (for now in catalog, productView and breadcrumbs stories)

components/src/atoms/breadcrumbs/Breadcrumbs.html
```diff
@@ -0,0 +1,35 @@
+<div class="breadcrumbs">
+  <template v-for="(breadcrumb, i) in breadcrumbs">
+    <template v-if="breadcrumbs.length === i + 1">
+      <!-- @slot Current page (Scoped slot) -->
+      <slot
+        name="current"
+        :breadcrumb="breadcrumb"
+      >
+        <span
+          :key="breadcrumb.text"
```
Aekal: Text of breadcrums will not always be unique, e.g. we can have the same name for category and product. It's the same case as here https://github.com/SnowdogApps/alpaca-ui/pull/161#discussion_r276017228. We should add id to breadcrumbs mock or use index with some prefix `:key="breadcrumb-${i}"`

components/src/atoms/breadcrumbs/Breadcrumbs.scss
```diff
@@ -0,0 +1,40 @@
+@import '../../../assets/styles/globals';
+
+$breadcrumbs__font-size        : $font-size-small !default;
+$breadcrumbs__item-color       : $blue !default;
+$breadcrumbs__link-color       : $gray-darker !default;
+$breadcrumbs__link-color-hover : $blue !default;
+$breadcrumbs__link-margin-after: 0 $spacer--medium !default;
```
Aekal: Use `--hover` and `--after` modifiers above

components/src/atoms/breadcrumbs/Breadcrumbs.html
jszczech: I will use `:key="breadcrumb-${i}"` for now ;)

components/src/04-views/catalog/Catalog.stories.js
Aekal: Right now the mock is in `Breadcrumbs` component folder, I think it'll be better to move it into global mocks in `components/mocks/`

PR #164, opened 2019-04-24
Feature/51274 - Discount code in minicart

Aekal:


components/src/03-modules/mini-cart/MiniCart.js
```diff
@@ -35,6 +37,27 @@ export default {
     taxAmount: {
       type: [String, Number],
       default: 0
+    },
```
robwozniak: Lets move `removeButton` props below all required to keep an order from the other components :slightly_smiling_face:

components/src/03-modules/modal/Modal.stories.js
```diff
@@ -29,9 +29,7 @@ storiesOf('Modules/Modal', module)
           <p>Lorem ipsum dolor sit amet, consectetur adipLorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.iscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.</p>
           <a-input
             label="Inline text"
-            type="text"
             id="field_id"
-            name="inline"
```
robwozniak: Are we sure name is not needed here? Maybe other (more appropriate one) could be ok? (same question for other deletions over this PR)

Aekal: `name` attribute in `input` is only for sending form data purposes, so it's not needed in these  components, especially on storybook level. It's safe to remove it, I didn't spot any specific logic related to these names. Anyway I'm going to refactor `input` component soon so I'll take a closer look at every case ;)

Aekal: By the way I spotted that every `a-button` after someone's refactoring didn't have `native` with `@click` event and because of that every click didn't work, e.g. modal. Fixed in last commit

robwozniak: Ok, great! :+1:

PR #166, opened 2019-04-24
Knobs addon

szafran89:


components/src/atoms/button/Button.stories.js
```diff
@@ -1,94 +1,60 @@
...
-  #### Use these modifiers with \`.button\` class.
+  ###### Check **Knobs** tab to edit component properties dynamically. Below list of available BEM modifiers.
   - \`.button--secondary\` - Selector for applying secondary styles
   - \`.button--blank\` - Selector for applying blank styles
   - \`.button--fluid\` - Selector for applying fluid styles (width: 100%)
   ---
 `
 
+const bemModifiers = [
+  'null',
```
Aekal: Change to `null` without quotes, because right now null is rendered as a class

components/src/atoms/button/Button.stories.js
```diff
@@ -1,94 +1,60 @@
...
+]
+
 storiesOf('Atoms/Button', module)
   .addParameters({ info })
-  // @vue/component
   .add('Default', () => ({
     components: { AButton },
+    props: {
+      customClass: {
+        default: select('BEM Modifier', bemModifiers, 'null')
```
Aekal: It looks like the first option from `bemModifiers` is set as default value, so we don't need to specify `'null'` here, you can remove it

components/src/atoms/button/Button.stories.js
```diff
@@ -1,94 +1,60 @@
...
       </a-button>
     `
   }))
-  // @vue/component
-  .add('Icon and text', () => ({
+  .add('With slot', () => ({
     components: { AButton, AIcon },
+    props: {
+      customClass: {
+        default: select('BEM Modifier', bemModifiers, 'null')
```
Aekal: Remove `'null'`

components/src/atoms/button/Button.stories.js
```diff
@@ -1,94 +1,60 @@
...
+const bemModifiers = [
+  'null',
+  'button--secondary',
+  'button--blank',
+  'button--fluid'
+]
+
 storiesOf('Atoms/Button', module)
   .addParameters({ info })
-  // @vue/component
```
Aekal: Afaik we need this comment for eslint

components/.storybook/config.js
```diff
@@ -4,12 +4,14 @@ import { configure, addDecorator, addParameters } from '@storybook/vue'
 // Addons
 import { withInfo } from 'storybook-addon-vue-info'
 import { withA11y } from '@storybook/addon-a11y'
+import { withKnobs } from "@storybook/addon-knobs";
```
jszczech: Semicolon ;)

components/src/atoms/button/Button.stories.js
szafran89: It doesn't make sense to lint this part of code. It show error when I don't set props types.

PR #172, opened 2019-05-06
Feature/53157 - UI - link

jszczech:


components/src/atoms/link/Link.html
```diff
@@ -1,3 +1,3 @@
-<a class="link">
+<a class="a-link">
   <slot />
```
Aekal: Add slot description for info tab

components/src/atoms/link/Link.scss
```diff
@@ -10,7 +10,7 @@ $link__transition           : $transition-base !default;
 $link__color--inverted      : $gray-darker !default;
 $link__color-hover--inverted: $blue !default;
```
Aekal: Please assign variables from `&--inverted` styles to global variables as in line above

Aekal: Sorry, I had something else in mind :D In line 35 for `&--secondary` we have these styles:
```
&--secondary {
    color: $gray-lighter;

    &:hover,
    &:focus {
      color: $color-primary;
    }
  }
```
please assign these colors to new variables to have them defined on the top

PR #175, opened 2019-05-07
#53158 - UI - list, list-item

jszczech:


components/src/atoms/list-item/ListItem.spec.js
```diff
@@ -0,0 +1,37 @@
...
+    expect(wrapper.is('li')).toBe(true)
+    expect(wrapper.classes()).toContain('a-list-item')
+    expect(wrapper.classes().length).toBe(1)
+  })
+
+  it('renders slot text when passed', () => {
+    const wrapper = mount(AListItem, {
+      slots: {
+        default: `
+          <span>List item default text</span>
```
Aekal: You can put it in one line
```
default: '<li>List default text</li>'
```

components/src/atoms/list-item/ListItem.spec.js
```diff
@@ -0,0 +1,37 @@
...
+        `
+      }
+    })
+
+    expect(wrapper.find('.a-list-item span').exists()).toBe(true)
+    expect(wrapper.find('.a-list-item span').text()).toEqual('List item default text')
```
Aekal: I think you can change selectors to `'.a-list-item > span'` to make sure that `span` is a direct child

components/src/atoms/list/List.spec.js
```diff
@@ -0,0 +1,37 @@
...
+    const wrapper = mount(AList, {
+      slots: {
+        default: `
+          <li>List default text</li>
+        `
+      }
+    })
+
+    expect(wrapper.find('.a-list li').exists()).toBe(true)
+    expect(wrapper.find('.a-list li').text()).toEqual('List default text')
```
Aekal: `.a-list > li` selector

PR #176, opened 2019-05-09
#53154

szafran89:


components/src/atoms/fieldset/Fieldset.spec.js
```diff
@@ -0,0 +1,31 @@
+import { mount } from '@vue/test-utils'
+import AFieldset from './Fieldset.vue'
+
+describe('Fieldset', () => {
+  it('has default structure', () => {
+    const wrapper = mount(AFieldset, {
+      propsData: {
+        legendText: 'Legend text'
```
jszczech: You can check the value of props also :)

PR #179, opened 2019-05-13
Feature/53294 - Tabs a11y

jszczech:


components/src/molecules/tabs/Tabs.js
```diff
@@ -22,8 +24,41 @@ export default {
     selectTab (selectedTab) {
       this.$emit('click', selectedTab.name)
       this.tabs.forEach(tab => {
-        tab.isActive = (tab.name === selectedTab.name)
+        tab.isActive = selectedTab === this.activeFocusedTab
+          ? tab.name === this.tabs[this.activeFocusedTab].name
+          : tab.name === selectedTab.name
```
Aekal: What do you think about changing this condition to:
```
if (selectedTab === this.activeFocusedTab) {
  tab.isActive = tab.name === this.tabs[this.activeFocusedTab].name
} else {
  tab.isActive = tab.name === selectedTab.name
}
```
for better readibility, it's hard to understand this shorthand ;)

components/src/molecules/tabs/Tabs.js
```diff
@@ -22,8 +24,43 @@ export default {
...
+    handleEvt (evt) {
+      const key = evt.keyCode
+
+      if (key === KeyCodes.TAB) {
+        evt.preventDefault()
+        evt.stopPropagation()
+        this.$refs.content.focus()
+      }
+
+      if ((key === KeyCodes.LEFT || key === KeyCodes.UP || key === KeyCodes.HOME) && this.activeFocusedTab > 0) {
```
anqaka: my only concern: shouldn't we use here `else if` instead of separate `if` condition?

components/src/molecules/tabs/Tabs.js
```diff
@@ -22,8 +24,39 @@ export default {
...
+      this.$refs[el][0].focus()
+    },
+    handleEvt (evt) {
+      const key = evt.keyCode
+
+      if (key === KeyCodes.TAB) {
+        evt.preventDefault()
+        evt.stopPropagation()
+        this.$refs.content.focus()
+      } else if ((key === KeyCodes.LEFT || key === KeyCodes.UP || key === KeyCodes.HOME) && this.activeFocusedTab > 0) {
```
szafran89: please split else if for multiple lines

components/src/molecules/tabs/Tabs.html
```diff
@@ -17,6 +18,7 @@
     type="button"
     role="tab"
     @click="selectTab(tab)"
+    @keydown="handleEvt"
```
szafran89: maybe better is rename to `onKeydown`

components/src/molecules/tabs/Tabs.scss
```diff
@@ -36,12 +37,21 @@ $tabs__content-padding\@large   : $spacer--large $spacer--extra-large !default;
       background-color: $tabs__button-background--active;
       border-bottom: $tabs__button-border--active;
     }
+
+    &:focus {
+      z-index: 1;
+      outline: $tabs__outline-focus;
+    }
```
szafran89: Can we show focus only for keyboard users? Lately, I talked with Ania there is some solution. As I remember it works for firefox but there is also a polyfill. Please ask Ania about details. If it works we should add it globally not only for tabs.

components/src/molecules/tabs/Tabs.js
```diff
@@ -22,8 +24,39 @@ export default {
...
+          tab.isActive = tab.name === this.tabs[this.activeFocusedTab].name
+        } else {
+          tab.isActive = tab.name === selectedTab.name
+        }
       })
+    },
+    focus (el) {
+      this.$refs[el][0].focus()
+    },
+    handleEvt (evt) {
```
szafran89: evt -> e

PR #180, opened 2019-05-14
#53598 - split Radio component to Radio and RadioWrapper

szafran89:


components/src/atoms/radio/Radio.html
```diff
@@ -0,0 +1,26 @@
+
```
Aekal: Empty line

components/src/atoms/radio/Radio.html
```diff
@@ -0,0 +1,26 @@
...
+<div class="a-radio">
+  <input
+    :name="name"
+    :id="id"
+    :value="value"
+    :checked="checked === value ? true : false"
```
Aekal: Use shorthand `:checked="checked === value"`

components/src/atoms/radio/Radio.js
```diff
@@ -19,24 +23,24 @@ export default {
...
     /**
-     * Class for input
+     * Input value
      */
-    inputClass: {
-      type: String,
-      default: null
+    value: {
+      type: [String, Number, Boolean],
+      required: true
```
Aekal: I think in some cases value will be the same as `id`, maybe let's change `value` to not be required and use `id` in this case. 
Then instead of
```
<a-radio
  v-model="selected"
  name="radio1"
  id="option_one"
  value="option_one"
>
  Option one
</a-radio>
```
we can use
```
<a-radio
  v-model="selected"
  name="radio1"
  id="option_one"
>
  Option one
</a-radio>
```
What do you think?

components/src/molecules/radio-wrapper/RadioWrapper.stories.js
```diff
@@ -0,0 +1,59 @@
...
+          name="radio"
+          v-model="selected"
+        >
+          Option two
+        </a-radio>
+        <a-radio
+          id="id-3"
+          value="option_three"
+          name="radio"
+          v-model="selected"
```
Aekal: I saw in vuetify and element support for `v-model` in radio-wrapper. In group of radio they specify one `v-model` in wrapper instead in each radio component https://codepen.io/pen/?&editable=true. Do you think we can do it too?

components/src/atoms/radio/Radio.html
szafran89: removed

components/src/atoms/radio/Radio.html
szafran89: done

components/src/atoms/radio/Radio.js
szafran89: added computed inputId 
```
  computed: {
    inputId () {
      return this.id || this.value.toLowerCase().replace(/[^a-z0-9]/g, '-')
    }
  }
```

components/src/molecules/radio-wrapper/RadioWrapper.stories.js
szafran89: I will create an issue for that

PR #183, opened 2019-05-15
Feature/53662

szafran89:


components/src/atoms/password-strength/PasswordStrength.spec.js
```diff
@@ -0,0 +1,54 @@
...
+
+describe('PasswordStrength', () => {
+  it('has default structure', () => {
+    const wrapper = mount(APasswordStrength)
+    const $passwordIndicator = wrapper.find('.a-password-strength__indicator')
+
+    expect(wrapper.is('div')).toBe(true)
+    expect(wrapper.classes()).toContain('a-password-strength')
+    expect(wrapper.classes().length).toBe(1)
+    expect($passwordIndicator.attributes('aria-live')).toBeDefined()
```
robwozniak: Is it needed if `toEqual` will return false if value will be undefined - empty attribute?

PR #184, opened 2019-05-16
#52364 - Add product list item component

robwozniak:


components/src/molecules/product-list-item/ProductListItem.stories.js
```diff
@@ -0,0 +1,96 @@
...
+        price=""
+        specialPrice="$100,00"
+        oldPrice="$299,99"
+        url="#"
+        :options="options"
+        @remove="removeMethod"
+      />
+    `
+  }))
+  .add('With Qty', () => ({
```
jszczech: Imo we dont need this story. Instead of that we should have story 'With slots', where we use all slots (also with quantity-update). 
The second point is: maybe we should add default quantity-update inside the component and then overwrite it with named slots ?

components/src/molecules/product-list-item/ProductListItem.html
```diff
@@ -0,0 +1,94 @@
...
+        'product-list-item__actions',
+        { 'product-list-item__actions--to-right': !$slots.quantity }
+      ]"
+    >
+      <div
+        v-if="$slots.quantity"
+        class="product-list-item__qty-wrapper"
+      >
+        <!--@slot Custom product quantity (Named slots)-->
+        <slot name="quantity"/>
```
jszczech: Imo inside this named slots we should have default quantity update element with possibility to hide.

components/src/molecules/product-list-item/ProductListItem.stories.js
```diff
@@ -0,0 +1,96 @@
...
+    ...defaultData,
+    methods: {
+      removeMethod: action('Remove'),
+      changeMethod: action('Change')
+    },
+    template: `
+      <a-product-list-item
+        name="Ingrid Running Jacket"
+        imageUrl="/images/product/cart-product-160_160.jpg"
+        price=""
```
jszczech: This price should be empty?

components/src/molecules/product-list-item/ProductListItem.js
```diff
@@ -0,0 +1,97 @@
...
+      type: Array,
+      default: null
+    }
+  },
+  computed: {
+    /**
+     * Dynamically generated item id based on name and timestamp
+     */
+    id () {
+      return `${new Date().getTime()}_${String(this.name).toLowerCase().replace(/\s/g, '-')}`
```
jszczech: Do we need it? Imo every product should has id in data :)

components/src/molecules/product-list-item/ProductListItem.scss
```diff
@@ -0,0 +1,149 @@
...
+    }
+  }
+
+  &__qty {
+    display: flex;
+  }
+
+  &__qty-label {
+    align-self: center;
+    margin: 0 $spacer 0 0;
```
jszczech: Move it to the variables.

components/src/molecules/product-list-item/ProductListItem.stories.js
```diff
@@ -0,0 +1,96 @@
...
+import AQuantityUpdate from '../../molecules/quantity-update/QuantityUpdate.vue'
+
+const info = 'Product List Item component'
+
+// @vue/component
+const defaultData = {
+  components: {
+    ALabel,
+    AProductListItem,
+    AQuantityUpdate
```
jszczech: First story use only one from this components, so we should separate it ;)

components/src/molecules/product-list-item/ProductListItem.html
```diff
@@ -0,0 +1,94 @@
+<a-list-item tag="div" class="product-list-item">
```
jszczech: Use a- prefix in class

components/src/molecules/product-list-item/ProductListItem.stories.js
robwozniak: Yes, this is required props so as long as I am not working on price element I would not change it under this task. Let me know if we should adjust this component anyway :slightly_smiling_face:

components/src/molecules/product-list-item/ProductListItem.js
robwozniak: You're right, moved to props.

components/src/molecules/product-list-item/ProductListItem.stories.js
robwozniak: Ok, I replace it with `Without qty` scenario, also qty named slot was updated. Please check and let me know what do you think. thanks

components/src/molecules/product-list-item/ProductListItem.html
```diff
@@ -0,0 +1,114 @@
...
+        'a-product-list-item__actions',
+        { 'a-product-list-item__actions--to-right': hideQty }
+      ]"
+    >
+      <div
+        v-if="!hideQty"
+        class="a-product-list-item__qty-wrapper"
+      >
+        <!--@slot Custom product quantity (Named slots)-->
+        <slot name="quantity">
```
Aekal: Change to scoped slot with `onChange` and `qty`

components/src/molecules/product-list-item/ProductListItem.html
```diff
@@ -0,0 +1,114 @@
...
+                Qty:
+              </a-label>
+            </template>
+          </a-quantity-update>
+        </slot>
+      </div>
+
+      <div class="a-product-list-item__action-buttons-wrapper">
+        <!--@slot Custom product remove button (Named slots)-->
+        <slot name="removeAction">
```
Aekal: Change to scoped slot with `onRemove`.
After all the above you can update `With slot` story where `<template>` is used and change the values to these from scope instead of defining them 2 times

components/src/molecules/product-list-item/ProductListItem.js
```diff
@@ -0,0 +1,114 @@
...
+     * Specify if qty component is not needed (default: false)
+     */
+    hideQty: {
+      type: Boolean,
+      default: false
+    }
+  },
+  methods: {
+    /**
+     * Triggered when value is removed by id
```
Aekal: I think you need to move these event comments one line above `this.$emit` to be rendered

robwozniak: You're right, thanks - fixed!

src/molecules/product-list-item/ProductListItem.html
```diff
@@ -0,0 +1,128 @@
+<a-list-item tag="div" class="a-product-list-item">
+  <div class="a-product-list-item__image">
+    <!--@slot Custom product image (Scoped slot)-->
+    <slot
+      name="image"
+      :alt="name"
+      :image-url="imageUrl"
```
szafran89: @e have to keep the same approach. If here we have `image-url` for  `name` slot we shoud also have `name` and `url` without prefixes

src/molecules/product-list-item/ProductListItem.html
```diff
@@ -0,0 +1,128 @@
...
+            decrement-icon-title="Minus mark"
+            increment-aria-label="Increase the quantity"
+            increment-icon-title="Plus mark"
+          >
+            <template #label>
+              <a-label
+                class="product-list-item__qty-label"
+                for="qty_1"
+              >
+                Qty:
```
szafran89: What about translations. we have to move it outside of component template.

src/molecules/product-list-item/ProductListItem.html
```diff
@@ -0,0 +1,128 @@
...
+        tag="dl"
+        class="a-product-list-item__attributes"
+      >
+        <template v-for="item in options">
+          <!-- @slot Custom product item option (Scoped slot) -->
+          <slot
+            name="option"
+            :option="item"
+          >
+            <a-list-item tag="dt">
```
szafran89: It looks different on design 
image
Please think about adding styles for a descriptive list, maybe it makes sense to add them as list variant?

src/molecules/product-list-item/ProductListItem.html
```diff
@@ -0,0 +1,128 @@
...
+        v-if="!hideQty"
+        class="a-product-list-item__qty-wrapper"
+      >
+        <!--@slot Custom product quantity (Scoped slot)-->
+        <slot
+          name="quantity"
+          :item-qty="qty"
+          :change-qty="onChange"
+        >
+          <a-quantity-update
```
szafran89: What do you think about using quantity as not default element? the same as here https://github.com/SnowdogApps/alpaca-ui/blob/develop/src/03-modules/product-item/ProductItem.html

Add just quantity slot and use if needed

szafran89: and also please check styles b/c qty label is on the left. I think we can add default variant story, cart variant, with slots at least

src/molecules/product-list-item/ProductListItem.html
```diff
@@ -0,0 +1,128 @@
...
+          name="removeAction"
+          :item-remove="onRemove"
+        >
+          <a-button
+            class="a-product-list-item__action-button"
+            aria-label="Remove this product from your shopping cart"
+            custom-class="a-product-list-item__actions-item"
+            @click.native="onRemove"
+          >
+            <a-icon
```
szafran89: wrong icon color

src/molecules/product-list-item/ProductListItem.js
```diff
@@ -0,0 +1,95 @@
...
+      type: String,
+      required: true
+    },
+    /**
+     * Product old price
+     */
+    oldPrice: {
+      type: String,
+      required: true
+    },
```
szafran89: IMO it would be better to pass one object price as storefront did here https://github.com/DivanteLtd/storefront-ui/blob/master/packages/vue/src/components/molecules/SfProductCard/SfProductCard.html

src/molecules/product-list-item/ProductListItem.js
```diff
@@ -0,0 +1,95 @@
...
+      type: Array,
+      default: null
+    },
+    /**
+     * Product quantity
+     */
+    qty: {
+      type: Number,
+      default: 1
+    },
```
szafran89: if we remove qty from default structure we can also remove this `qty` and `hideQty` props

src/molecules/product-list-item/ProductListItem.scss
```diff
@@ -0,0 +1,147 @@
...
+  &__data {
+    display: flex;
+    flex-flow: row wrap;
+  }
+
+  &__name {
+    flex: 1;
+    margin-bottom: $spacer;
+    line-height: $font-line-height;
+    text-decoration: none;
```
szafran89: move to local var

src/molecules/product-list-item/ProductListItem.scss
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  &__qty {
+    display: flex;
+  }
+
+  &__qty-label {
+    align-self: center;
+    margin: $product-list-item__qty-label-margin;
+    font-weight: normal;
```
szafran89: move to local var

src/molecules/product-list-item/ProductListItem.stories.js
```diff
@@ -0,0 +1,131 @@
+import { storiesOf } from '@storybook/vue'
+import { action } from '@storybook/addon-actions'
+
+import AProductListItem from './ProductListItem.vue'
+
+const info = 'Product List Item component'
+
+// @vue/component
```
szafran89: we can remove this comment in case of stories file

src/molecules/product-list-item/ProductListItem.stories.js
```diff
@@ -0,0 +1,131 @@
...
+        }
+      ]
+    }
+  },
+  methods: {
+    removeMethod: action('Remove'),
+    changeMethod: action('Change')
+  }
+}
+
```
szafran89: Can we add some knobs?

src/molecules/product-list-item/ProductListItem.html
robwozniak: Ok, now it's going as a slot in stories so it's independent. Let me know if we need still to improve this translation.

src/molecules/product-list-item/ProductListItem.html
robwozniak: You're right, it was related with some other changes in child components. Styles were added and used in this component, let me know what do you think :slightly_smiling_face:

src/molecules/product-list-item/ProductListItem.html
robwozniak: It's hard to get it from zeplin (also there is no clue in svg file) but imo this is base color with changed opacity. Styles updated, let me know if you would like to add new color here.

src/molecules/product-list-item/ProductListItem.html
robwozniak: Maybe it's better to move these variants directly to qty component? I mean add modifiers to specify different label positions - `--top`, `--left` etc

src/molecules/product-list-item/ProductListItem.stories.js
robwozniak: Sure, added for product: name, options, image url and price.

src/molecules/product-list-item/ProductListItem.html
szafran89: it looks a little strange if we have long labels 
image
I'll try to change it, I think we have to use unfortunately `float` property. the same as here https://alpaca-components.now.sh/components/detail/list--description.html

src/molecules/product-list-item/ProductListItem.html
szafran89: I'll check this

PR #190, opened 2019-05-30
Feature/53855 Migrate footer component

jszczech:


components/src/organisms/footer/Footer.js
```diff
@@ -16,48 +17,94 @@ export default {
...
+    }
+  },
+  watch: {
+    windowWidth (newWidth) {
+      this.isLargeScreen = newWidth > 768
     }
   },
+  mounted () {
+    this.$nextTick(() => {
+      window.addEventListener('resize', () => {
```
Aekal: It's not good for performance, please consider using [debounce](http://bencentra.com/code/2015/02/27/optimizing-window-resize.html) or matchMedia (Robert told us last time matchmedia is optimized but it needs to be checked).
Same in the watcher above

components/src/organisms/footer/Footer.html
```diff
@@ -0,0 +1,116 @@
+<footer class="a-footer">
+  <div class="a-footer__handler container">
+    <!-- @slot Scroll to top button (Scoped slot) -->
+    <slot
+      name="scrollButton"
+      :scroll-to-top="scrollToTop"
+    >
+      <a-button
+        class="a-footer__scroll-top"
+        aria-label="Scroll to top"
```
Aekal: Use prop

components/src/organisms/footer/Footer.html
```diff
@@ -0,0 +1,116 @@
...
+        @click.native="scrollToTop"
+      >
+        <a-icon
+          icon="arrow-up"
+          class="a-footer__icon"
+        />
+      </a-button>
+    </slot>
+    <section class="a-footer__newsletter col-xs-12 col-lg-3">
```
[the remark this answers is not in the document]
jszczech: after changes in structure it's no needed ;)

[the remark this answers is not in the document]
Igloczek: Do you think that is there a way to use the grid as a component? Now it kinda depend on some not defined dependency, so you can't use this component separately.

components/src/molecules/collapse/Collapse.html
```diff
@@ -7,10 +7,11 @@
     :ariaControls="ariaControls"
   >
     <a-button
-      class="a-collapse__button a-button--blank"
+      :class="['a-collapse__button a-button--blank', buttonClass]"
```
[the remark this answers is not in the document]
jszczech: In this case we want to add a little changes, all the rest should be the same. 
I'm not sure if duplicating styles a second time is a good idea. We could add style for `a-collapse__button` in footer, but imo it's wrong idea. What do you think? Do we need override entire button here for sure?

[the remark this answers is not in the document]
Aekal: Ok let's stay with it for now

src/organisms/footer/Footer.html
```diff
@@ -0,0 +1,134 @@
...
+        :aria-label="scrollAriaLabel"
+        @click.native="scrollToTop"
+      >
+        <a-icon
+          icon="arrow-up"
+          class="a-footer__icon"
+        />
+      </a-button>
+    </slot>
+    <section class="a-footer__newsletter col-xs-12 col-lg-3">
```
szafran89: Can we remove grid classes and use only in pages? IMO we should try

src/organisms/footer/Footer.html
```diff
@@ -0,0 +1,134 @@
...
+    </div>
+  </section>
+  <!--TODO Edit after merge cookie-message component-->
+  <!-- @slot Cookie message (Named slot) -->
+  <slot name="cookie">
+    <a-cookie-message
+      :text="cookieText"
+      closeIcon="close"
+      dataType="cookie"
+      closeButton="Close"
```
szafran89: `close-button` and the same for other props in stories file

src/organisms/footer/Footer.scss
```diff
@@ -1,29 +1,34 @@
...
 $footer__scroll-top-top              : -24px !default;
 $footer__icon-size                   : 13px !default;
 $footer__icon-fill                   : $color-secondary !default;
 $footer__icon-fill-hover             : $color-primary !default;
 
-.footer {
+.a-footer {
   flex-shrink: 0;
-  background-color: $footer__background-color;
+  background-color: $footer__background-color-darkest;
```
szafran89: here `$footer__background-color`

src/organisms/footer/Footer.scss
```diff
@@ -1,29 +1,34 @@
...
 
   @include mq($screen-l) {
-    background-color: $footer__background-color\@large;
+    background-color: $footer__background-color-darker;
```
szafran89: `$footer__background-color\@large`

src/organisms/footer/Footer.scss
```diff
@@ -44,43 +49,75 @@ $footer__icon-fill-hover             : $color-primary !default;
   &__newsletter {
     margin: 0 auto $footer__padding;
```
szafran89: `$footer__padding` for margin? we should set new var

src/organisms/footer/Footer.scss
```diff
@@ -44,43 +49,75 @@ $footer__icon-fill-hover             : $color-primary !default;
...
-    .dropdown-list__item {
-      @include mq($screen-m) {
-        border-bottom: 0;
+    &-button {
+      margin-bottom: $footer__link-button-margin;
+      color: $footer__link-button-color;
+      font-weight: bold;
+      background-color: $footer__background-color-darker;
+
+      .a-icon {
```
szafran89: Can we use BEM class?

src/organisms/footer/Footer.scss
```diff
@@ -117,6 +154,10 @@ $footer__icon-fill-hover             : $color-primary !default;
       order: 1;
       margin: $reset;
     }
+
+    .a-list {
```
szafran89: can we add new class instead of this root class?

src/molecules/footer-menu/FooterMenu.html
```diff
@@ -0,0 +1,44 @@
...
+    :key="item.id"
+    :visible="isLargeScreen"
+    :title="item.title"
+    :id="item.id"
+  >
+    <template #button="{ visibility, ariaControls, toggle }">
+      <a-button
+        :class="[
+          'a-button--blank a-collapse__button a-footer-menu__button',
+          { ' a-footer-menu__button--active': visibility }
```
Aekal: Please remove extra space before 'a-footer-'

src/organisms/footer/Footer.html
```diff
@@ -1,134 +1,106 @@
...
-              <a-link
-                class="list__icon-link"
-                :href="item.link"
+    <a-container
+      tag="div"
+      class="a-footer__bottom-bar-container"
+    >
+
+      <!-- @slot Social media container (Named slot) -->
+      <slot name="socias">
```
Aekal: Are you sure we can nest slots inside slot?  Please check if it works when we use all slots at once, if not maybe just remove this parent slot `<slot name="socias">`

src/molecules/footer-menu/FooterMenu.html
```diff
@@ -0,0 +1,44 @@
...
+    class="a-footer-menu__list"
+    button-class="a-footer-menu__button"
+    content-class="a-footer-menu__content"
+    v-for="item in menu"
+    :key="item.id"
+    :visible="isLargeScreen"
+    :title="item.title"
+    :id="item.id"
+  >
+    <template #button="{ visibility, ariaControls, toggle }">
```
jszczech: IE alert 😃 

For now we need to avoid this approach :)

jszczech: Example

https://github.com/SnowdogApps/alpaca-ui/blob/f6b8e7a987fe91fe261ac3069201a4fc112ccc32/src/molecules/quantity-update/QuantityUpdate.stories.js#L100

szafran89: good point ;)

src/molecules/footer-menu/FooterMenu.html
szafran89: done

src/organisms/footer/Footer.html
szafran89: parent slot removed

PR #195, opened 2019-06-03
Structure refactor

Igloczek:
All components are now stored flatter, without split to components and cookbook, under single Storybook instance.

"Cookbook" components become now "template" type of component - it doesn't ideally fit into Atomic Design, but it's better than separate "cookbook" think, which is not self-explanatory.

All components have improved inheritance, by moving child components imports from JS to Vue files (I just wrote a [script for this](https://gist.github.com/Igloczek/b5b313cbcdeef6922f665568d88f00fd), so didn't test every component separately).

The JS part is now always used as a mixin, which makes it easier to extend in the future and solves some issues with scoped styles (it creates a new component instance).

There is also a new build process, which is way simpler than previous, it just generates a `index.js` which imports and exports all components, so it's now possible to import components like this:

```js
import { AButton } from '@snowdog/alpaca-ui'`
```

but when more flexibility is needed you can take the source:

```vue
<template src="@snowdog/alpaca-ui/src/atoms/button/Button.html" />
<script>
import Button from '@snowdog/alpaca-ui/src/atoms/button/Button.js'

export default {
  mixins: [Button]
}
</script>
<style src="@snowdog/alpaca-ui/src/atoms/button/Button.scss" lang="scss" /> 
```


src/01-globals/information/Information.vue
```diff
@@ -0,0 +1,18 @@
+<template src="./Information.html" />
+
+<script>
+import AHeading from '../../atoms/heading/Heading.vue'
+import AParagraph from '../../atoms/paragraph/Paragraph.vue'
+
+import Information from './Information.js'
```
szafran89: hmm maybe it would be better to add prefix also here  `AInformation` and the same proposal for other components imported like above.

PR #198, opened 2019-06-06
Feature/54625

szafran89:


src/atoms/table/Table.selectors.json
```diff
@@ -0,0 +1,6 @@
+[
+  {
+    "name": ".table--odd-even",
```
robwozniak: We don't have any styles for it, are you missing commit something or we're waiting for updated component?

src/atoms/table/Table.stories.js
```diff
@@ -25,7 +21,9 @@ const defaultData = {
 }
 
 storiesOf('Atoms/Table', module)
-  .addParameters({ info })
+  .addParameters({
+    info: generateVueInfoTable(selectorsConfig, 'BEM modifiers')
+  })
   .add('Default', () => ({
     ...defaultData,
```
robwozniak: `Table.selectors.json` contain one class but I can't see props with `classKnobs` regarding possible BEM class options.

src/molecules/message/Message.stories.js
```diff
@@ -1,17 +1,15 @@
...
-  #### Use these modifiers with \`.message\` class.
-  - \`.message--success\` - Selector for applying styles for success message
-  - \`.message--error\` - Selector for applying styles for error message
-  ---
+  ${generateVueInfoTable(selectorsConfig, 'BEM modifiers')}
 `
 
-// @vue/component
 const defaultData = {
   components: { AMessage },
```
robwozniak: We need to add props `classKnobs` and pass options from `Message.selectors.json` to make it visible in storybook app, now they are not available to choose :cry:

szafran89:
@robwozniak for heading and price we have different tasks created.

src/atoms/table/Table.selectors.json
szafran89: Related styles are under `TableRow.scss`

src/atoms/table/Table.stories.js
szafran89: added

src/molecules/message/Message.stories.js
szafran89: done

PR #204, opened 2019-06-12
feature/54836 - Update Container component

robwozniak:


szafran89:
Can we add knobs for tag? 
I also wonder if better is to use `main` or `div` as prop default value.

PR #205, opened 2019-06-12
feature/54505 - Update Paragraph class name

robwozniak:


szafran89:
Please add knobs for tag and text

szafran89:
Please update readme, remember about it in future ;)

PR #208, opened 2019-06-14
Feature/54508 Radio

jszczech:


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Added
 - Vue Info table generator helper
+- Tests for radio button
 
 ### Changed
 - heading: update heading class and add unit test and knobs
```
szafran89: Add note about style refactor as well

src/atoms/radio/Radio.spec.js
```diff
@@ -0,0 +1,96 @@
+import { mount } from '@vue/test-utils'
+import ARadio from './Radio.vue'
+
+describe('Radio', () => {
+  it('has default structure', () => {
+    const wrapper = mount(ARadio, {
+      propsData: {
+        id: '1',
```
szafran89: id not required so it's not part of default structure. Maybe we should set it as `required`?

src/molecules/radio/Radio.js
```diff
@@ -1,6 +1,10 @@
 // @vue/component
+import ALabel from '../../atoms/label/Label.vue'
```
szafran89: You can not add import in js file. As you can see in other components we add a child component to vue file to avoid inheritance  issue in the parent component.

szafran89: example here https://github.com/SnowdogApps/alpaca-ui/blob/develop/src/molecules/radio-wrapper/RadioWrapper.vue

PR #211, opened 2019-06-18
feature/54496 - Image component updates

robwozniak:


src/atoms/image/Image.scss
```diff
@@ -3,13 +3,13 @@
...
-.image {
+.a-image {
   display: block;
   width: $image__width;
   max-width: $image__width;
 }
 
 img,
 picture {
-  @extend .image;
```
jszczech: I think, we should remove extends, we don't need it.

PR #212, opened 2019-06-18
#54515 - Checkbox

jszczech:


src/molecules/checkbox/Checkbox.spec.js
```diff
@@ -0,0 +1,101 @@
+import { mount } from '@vue/test-utils'
+import ACheckbox from './Checkbox.vue'
+
+describe('Checkbox', () => {
+  it('has default structure', () => {
+    const wrapper = mount(ACheckbox, {
+      propsData: {
+        id: '1',
+        name: 'Checkbox',
+        value: 'checkbox'
```
szafran89: Shouldn't `id`, `name` and `value` be required props?

PR #213, opened 2019-06-19
#54499 - Input/ Input wrapper

jszczech:


src/molecules/input-wrapper/InputWrapper.spec.js
```diff
@@ -22,4 +22,10 @@ describe('Input wrapper', () => {
     const input = wrapper.find('.a-input-wrapper > input')
     expect(input.exists()).toBe(true)
   })
+
+  it('renders text input by default', () => {
+    const wrapper = mount(AInputWrapper)
+    const input = wrapper.find('.a-input')
+    expect(input.attributes('type')).toBe('text')
+  })
```
szafran89: Do we need this test? It's part of `Input.spec.js`

jszczech: Do you mean "renders text input by default" ? In input component we didn't have type, in input-wrapper - yes. But ... as I think now, maybe we should add props with type in input-wrapper component?

jszczech: Ok, after changes with default slot, we don't need it. Removed :D

PR #221, opened 2019-07-03
feature/54520 - Modal

robwozniak:


src/molecules/modal/Modal.stories.js
```diff
@@ -0,0 +1,256 @@
...
+      }
+    },
+    template: `
+      <div>
+        <a-button @click.native="showModal">
+          Modal button
+        </a-button>
+        <a-modal
+          name="default"
+          heading="Sample title"
```
Aekal: Put `:heading="headingKnob"` here

src/molecules/modal/Modal.stories.js
```diff
@@ -0,0 +1,256 @@
...
+  }))
+  .add('Without header', () => ({
+    components: {
+      AButton,
+      AModal,
+      AInput,
+      AInputWrapper
+    },
+    props: {
+      headingKnob: {
```
Aekal: This knob is not needed for this story 'without header'

src/molecules/modal/Modal.html
```diff
@@ -0,0 +1,64 @@
...
+            <a-divider
+              v-if="heading"
+              class="a-modal__divider"
+            />
+          </slot>
+          <button
+            @click="hide(name)"
+            class="button button--icon a-modal__close-button"
+            type="button"
+          >
```
Aekal: Please add aria-label - you can see a violation in accessibility tab after modal open.
Probably you can use `<a-button>` with `<a-icon>`and update styles (cursor pointer is also missing)

src/molecules/modal/Modal.js
```diff
@@ -1,41 +1,68 @@
...
     },
-    maxWidth: {
-      type: Number,
-      default: null
+    /**
+     * Modal main transition effect name
+     */
+    modalMainTransition: {
+      type: String,
+      default: 'fade'
```
Aekal: Can we add this transition type and `modalContentTransition` to knobs and use dynamically? 
The knob can be a list of `fade`, `slide-down`, `slide-right`

src/molecules/modal/Modal.scss
```diff
@@ -84,20 +87,23 @@ $modal__close-button-icon-transition : $transition-base !default;
...
   }
 
   &__header {
     display: flex;
+    flex-wrap: wrap;
     justify-content: space-between;
     margin: $modal__header-margin;
   }
 
+  &__header &__divider {
```
Aekal: Can it be just `&__divider`?.
In line 125 please change `&-icon` to `&__close-button-icon`

src/molecules/modal/Modal.spec.js
```diff
@@ -0,0 +1,101 @@
...
+
+  it('renders slot text when passed', () => {
+    const wrapper = mount(AModal, {
+      propsData: {
+        name: 'Test modal'
+      },
+      slots: {
+        default: `<span data-test="content">Sample text</span>`,
+        heading: `<h1 data-test="heading">Sample text</h1>`,
+        close: `<button data-test="close">Sample text</button>`
```
Aekal: Let's change above texts to more descriptive like `Sample text`, `Sample heading`, `Sample button`

src/molecules/modal/Modal.stories.js
robwozniak: Damn, you're right - thanks!

src/molecules/modal/Modal.js
robwozniak: Good idea, added.

src/molecules/modal/Modal.scss
robwozniak: Styles presented in `&__header &__divider` are for specific - default - header presentation case. I'll move them directly to divider as a BEM modifier. Let me know if you have any other idea for it.

Also, icon styles have been modified to bit simpler `&__close-icon`.

src/molecules/modal/Modal.html
Aekal: Aria-label is still missing :P

src/molecules/modal/Modal.stories.js
```diff
@@ -9,6 +9,13 @@ import AInput from '../../atoms/input/Input.vue'
 import AInputWrapper from '../../molecules/input-wrapper/InputWrapper.vue'
 import AModal from './Modal.vue'
 
+const availableTransitions = [
+  'fade',
+  'slide-down',
+  'slide-right',
+  'slide'
```
Aekal: It looks like the `slide` doesn't change anything, we can remove this one from the list

src/molecules/modal/Modal.stories.js
```diff
@@ -20,6 +27,12 @@ storiesOf('Molecules/Modal', module)
     props: {
       contentKnob: {
         default: text('Content', 'Lorem ipsum dolor sit amet, consectetur adipLorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.iscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.')
+      },
+      mainTransitionKnob: {
```
Aekal: I was telling you about adding these 2 knobs for transition but now I see that changing this type for the overlay is not that good, because only fade here makes sense. I think we can remove this one `mainTransitionKnob` and leave only `contentTransitionKnob` in all stories

src/molecules/modal/Modal.js
```diff
@@ -1,85 +1,100 @@
...
     },
     handleKeydown (event) {
-      if (this.visibility) {
+      if (this.visible) {
         switch (event.key) {
           case 'Esc': // IE/Edge specific value
           case 'Escape':
             if (this.closeOnEsc) {
-              this.hide(this.name)
+              this.$emit('close')
```
szafran89: Add to the separate method and add a comment to show event in Vue-info addon

src/molecules/modal/Modal.stories.js
```diff
@@ -0,0 +1,264 @@
...
+          @click.native="openModal()"
+        >
+          Modal button
+        </a-button>
+        <a-modal
+          :visible="isModalVisible"
+          :closeButton="false"
+          ref="modal"
+          @close="closeModal()"
+          :modalContentTransition="contentTransitionKnob"
```
szafran89: `modal-content-transition`

src/molecules/modal/Modal.stories.js
```diff
@@ -0,0 +1,264 @@
...
+      <div>
+        <a-button
+          ref="modalTriggerBtn"
+          @click.native="openModal()"
+        >
+          Modal button
+        </a-button>
+        <a-modal
+          :visible="isModalVisible"
+          :closeButton="false"
```
szafran89: `close-button`
check also other palaces and fix this rule

src/molecules/modal/Modal.js
```diff
@@ -1,85 +1,119 @@
...
       if (this.closeOnBackgroundClick) {
-        this.hide(this.name)
+        this.close()
       }
     },
     handleKeydown (event) {
-      if (this.visibility) {
+      if (this.visible) {
         switch (event.key) {
           case 'Esc': // IE/Edge specific value
```
szafran89: this case is empty, is it ok?

szafran89: in other modules we use `KeyCodes`. IMO we should follow the same approach

src/molecules/modal/Modal.js
```diff
@@ -1,85 +1,119 @@
...
-      return this.maxWidth !== null ? `max-width: ${this.maxWidth}px;` : null
+    ariaHidden () {
+      return this.visible ? 'false' : 'true'
     }
   },
-  methods: {
-    show (name) {
-      if (name === this.name) {
+  watch: {
+    visible (val) {
```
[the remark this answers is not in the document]
robwozniak: It's working well without this option, when `immediate` is added then on init process we don't have an access to `this` and what is more important to `this.trigger` data which is needed to indicate latest active DOM element

src/molecules/modal/Modal.js
robwozniak: It's empty because we want to use one action for two cases. Ok then, `keyCode` used instead `key`, as far as I checked it should be same kode over ie and other modern browsers.

src/molecules/modal/Modal.js
```diff
@@ -1,85 +1,119 @@
...
+      if (val && typeof window !== 'undefined') {
```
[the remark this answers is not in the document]
robwozniak: I am not sure but test env does not have an access to window, so for sure it should save us from issues during test process.

PR #227, opened 2019-07-08
#54506 - refactored Price component, added tests

szafran89:


src/atoms/price/Price.html
```diff
@@ -1,24 +1,28 @@
-<div class="price">
-  <!-- @slot Custom old price slots (Named slots) -->
+<div class="a-price">
+  <!-- @slot Custom old price (Named slots) -->
```
robwozniak: imo we can transform it to scoped slot and pass needed data, for example like `:special-price`, `:regular-price`. Please let me know what do you think, we can replace all named slots in this component to scoped.

src/atoms/price/Price.stories.js
```diff
@@ -2,77 +2,30 @@ import { storiesOf } from '@storybook/vue'
...
-  .add('Special Price', () => ({
-    ...defaultData,
-    template: `
-      <a-price
-        :special-price="special"
-        aria-label-special="Special price:"
-      />
-    `
+    components: { APrice },
+    template: `<a-price regular-price="$36,00" />`
```
robwozniak: Would be great to have some knobs (even if they are not super useful) to use in this story for some values :slightly_smiling_face:

szafran89: knobs added

src/atoms/price/Price.html
szafran89: as we discussed through the zoom

PR #229, opened 2019-07-14
Feature/54514 Select

jszczech:


src/molecules/select/Select.html
```diff
@@ -1,19 +1,22 @@
-<div class="select">
+<div class="a-select">
   <!-- @slot Label content (Named slot) -->
   <slot name="label">
```
robwozniak: Why not scoped? :wink:

src/molecules/select/Select.scss
```diff
@@ -1,50 +1,67 @@
...
+      right: $select__field-border-right--after;
+      margin: $select__field-margin--after;
+      border-color: $select__field-border-color--after;
+      border-style: $select__field-border-style--after;
+      border-width: $select__field-border-width--after;
       content: "";
-      pointer-events: none;
     }
+
+    &--icon {
```
robwozniak: Imo we should add bigger padding right on `__input` wrapper to avoid such situation:

src/molecules/select/Select.scss
```diff
@@ -1,50 +1,67 @@
...
+$select__field-border-radius      : 20px !default;
+$select__field-margin--after      : 3px 0 0 0;
+$select__field-border-color--after: $gray-dark transparent transparent transparent !default;
+$select__field-border-right--after: $spacer--medium !default;
+$select__field-border-style--after: solid !default;
+$select__field-border-width--after: 5px !default;
+$select__icon-right               : $spacer--medium !default;
+$select__icon-width               : 16px !default;
+$select__icon-height              : 16px !default;
+$select__input-height             : 40px !default;
```
robwozniak: Is it correct with design mockups?

src/molecules/select/Select.scss
```diff
@@ -1,50 +1,67 @@
...
-      margin-top: -2.5px;
-      border-color: $select__arrow-color;
-      border-style: solid;
-      border-width: 5px;
+      right: $select__field-border-right--after;
+      margin: $select__field-margin--after;
+      border-color: $select__field-border-color--after;
+      border-style: $select__field-border-style--after;
+      border-width: $select__field-border-width--after;
       content: "";
```
robwozniak: Let's move content prop to the top

src/molecules/select/Select.stories.js
```diff
@@ -12,44 +12,61 @@ storiesOf('Molecules/Select', module)
     data () {
```
robwozniak: Maybe we can move it to const and use here and in `Custom icon` story?

src/molecules/select/Select.html
jszczech: Make sense, thanks ! :blush:

jszczech: I removed `label="Label text"`from story `With slots` - now should be ok :blush:

src/molecules/select/Select.scss
jszczech: Nope, I fixed it :D

src/molecules/select/Select.scss
jszczech: It's stylelint formatting :blush: Should we change it? ;) 

image

robwozniak: Ok then, I am not sure, so please keep it as it is

src/molecules/select/Select.scss
```diff
@@ -1,50 +1,72 @@
 @import '../../../assets/styles/_globals.scss';
```
robwozniak: Imo we can add `__font-size` var with design/app friendly font-size, now inherited styles looks like this `font-size: 100%;` and it does not look really good.

src/molecules/select/Select.html
```diff
@@ -1,19 +1,22 @@
...
       v-if="label"
     >
       {{ label }}
     </a-label>
   </slot>
-  <div class="select__field">
+  <div :class="[
+      'a-select__field',
+      icon && 'a-select__field--icon'
+    ]">
```
Aekal: Indentation

jszczech: I moved it in one line, bc it's short :blush:

PR #231, opened 2019-07-15
feature/54474 - Breadcrumbs

robwozniak:


src/atoms/breadcrumbs/Breadcrumbs.html
```diff
@@ -1,41 +1,47 @@
...
-        name="current"
-        :breadcrumb="breadcrumb"
-      >
-        <span
-          :key="`breadcrumb-${i}`"
-          :aria-label="breadcrumb.text"
-          class="a-breadcrumbs__item"
-          tabindex="0"
+<nav aria-label="Breadcrumb" class="a-breadcrumbs">
+  <ol class="a-breadcrumbs__list">
```
jszczech: Why you didn't use list component?

src/atoms/breadcrumbs/Breadcrumbs.spec.js
```diff
@@ -4,14 +4,19 @@ import ABreadcrumbs from './Breadcrumbs.vue'
 
 describe('Breadcrumbs', () => {
   it('has default structure', () => {
```
jszczech: Maybe we should check if the last element is clickable?

src/atoms/breadcrumbs/Breadcrumbs.stories.js
```diff
@@ -40,7 +51,10 @@ storiesOf('Atoms/Breadcrumbs', module)
           </span>
         </template>
         <template #current="data">
-          <span style="padding: 0 10px; color: #2962ff; line-height: 1.5;">
+          <span
+            aria-current="page"
+            style="padding: 0 10px; color: #2962ff; line-height: 1.5;"
```
jszczech: Imo we should remove line-height in all places in this component, it could be problematic in the future :)

src/atoms/breadcrumbs/Breadcrumbs.html
```diff
@@ -1,41 +1,47 @@
...
-      <slot
-        name="current"
-        :breadcrumb="breadcrumb"
-      >
-        <span
-          :key="`breadcrumb-${i}`"
-          :aria-label="breadcrumb.text"
-          class="a-breadcrumbs__item"
-          tabindex="0"
+<nav aria-label="Breadcrumb" class="a-breadcrumbs">
```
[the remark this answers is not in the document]
robwozniak: It's tricky (breadcrumb translation) in most of the other languages to keep the component context, but you're right - added :smile:

src/atoms/breadcrumbs/Breadcrumbs.scss
```diff
@@ -1,35 +1,51 @@
...
+
+  &__list-item {
     display: flex;
+    align-items: center;
+  }
+
+  &__link {
     color: $breadcrumbs__link-color;
-    text-decoration: none;
+    line-height: $breadcrumbs__line-height;
```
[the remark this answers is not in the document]
robwozniak: It's from the old component version code, but I guess we don't - removed

src/atoms/breadcrumbs/Breadcrumbs.html
robwozniak: I want to keep it as an atom - complete independent from the other components

szafran89: IMO `List` component can provide unnecessary complexity. So I vote for omitting that component.

src/atoms/breadcrumbs/Breadcrumbs.spec.js
robwozniak: Imo is not needed right now, we are depend on `$breadcrumbs__link-pointer-events--current` which allows us to change it dynamically so I am not sure if we can check it on js level

src/atoms/breadcrumbs/Breadcrumbs.html
jszczech: I loosed :D

src/atoms/breadcrumbs/Breadcrumbs.scss
```diff
@@ -1,35 +1,50 @@
...
-$breadcrumbs__font-size        : $font-size-small !default;
-$breadcrumbs__line-height      : 1.5 !default;
-$breadcrumbs__item-color       : $blue !default;
-$breadcrumbs__link-color       : $gray-darker !default;
-$breadcrumbs__link-color--hover: $blue !default;
-$breadcrumbs__separator-margin : 0 $spacer !default;
+$breadcrumbs__list-margin                 : 0 !default;
+$breadcrumbs__list-padding                : 0 !default;
+$breadcrumbs__list-item-margin            : 0 0 $spacer 0 !default;
+$breadcrumbs__font-size                   : $font-size-small !default;
```
szafran89: move to top, keep the same ore as it's used in code

Additionally you can add an empty line between a group of vars, it's more readable
```
$breadcrumbs__font-size                   : $font-size-small !default;

$breadcrumbs__list-margin                 : 0 !default;
$breadcrumbs__list-padding                : 0 !default;

$breadcrumbs__list-item-margin            : 0 0 $spacer 0 !default;

$breadcrumbs__link-color                  : $black !default;
$breadcrumbs__link-color--hover           : $blue !default;
$breadcrumbs__link-color--current         : $blue !default;
$breadcrumbs__link-pointer-events--current: none !default;
$breadcrumbs__link-text-decoration        : none !default;

$breadcrumbs__separator-margin            : 0 $spacer !default;
$breadcrumbs__separator-color             : $breadcrumbs__link-color !default;
```

src/atoms/breadcrumbs/Breadcrumbs.stories.js
```diff
@@ -16,16 +17,26 @@ storiesOf('Atoms/Breadcrumbs', module)
...
     components: { ABreadcrumbs },
+    props: {
+      breadcrumbsKnob: {
+        default: object('Breadcrumbs links', breadcrumbs)
+      }
+    },
     ...defaultData,
     template: `
-      <a-breadcrumbs :breadcrumbs="breadcrumbs"/>
+      <a-breadcrumbs :breadcrumbs="breadcrumbsKnob"/>
```
szafran89: keep in one line

PR #233, opened 2019-07-17
#54499 - Add slot for icon in input component

jszczech:


src/molecules/input/Input.html
```diff
@@ -8,12 +8,15 @@
...
     v-bind="$attrs"
     v-on="listeners"
     :value="value"
     :id="id"
     class="a-input__field"
   >
+  <div class="a-input__icon">
+    <!-- @slot Icon (Named slot) -->
+    <slot name="icon" />
+  </div>
```
szafran89: Please add a new variant to show how we can get input with icon

szafran89: or update variant `with slot` and change the name to `with slots`

szafran89:
After icon click
http://prntscr.com/ogpg0t

szafran89:
something wrong with linter 
image

PR #235, opened 2019-07-18
Feature/56163 2

szafran89:


src/molecules/modal/Modal.html
```diff
@@ -1,19 +1,20 @@
...
+    :role="modalRole"
   >
     <div
       @click.self="handleBackgroundClick"
       class="a-modal__mask"
     ></div>
     <transition :name="modalContentTransition">
       <div
-        class="a-modal__content"
+        :class="['a-modal__content', modalContentClass]"
```
szafran89: do we need `modalContentClass` prop?

src/molecules/modal/Modal.selectors.json
```diff
@@ -2,5 +2,21 @@
   {
     "name": ".a-modal--blank",
     "description": "Remove default content paddings"
+  },
+  {
+    "name": ".a-modal--compact",
```
szafran89: do we need this modifier?

robwozniak: It was taken from the old code, I can't find in designs what was the reason of this modifier so I'll remove it - if needed in future then we can easily add appropriate styles on higher component level

src/molecules/modal/Modal.html
robwozniak: Imo it's beneficial but for future usage, now it's not really needed - not used anywhere

PR #244, opened 2019-08-29
feature/56543 - ProductListItem styles update

robwozniak:


src/molecules/product-list-item/ProductListItem.scss
```diff
@@ -87,9 +87,12 @@ $product-list-item__qty-label-margin            : 0 $spacer 0 0 !default;
       justify-content: flex-end;
     }
 
-    &--special,
-    &--old {
+    .a-price__old {
```
szafran89: You know I don't like selector like this under BEM element, Is there any other option to update these styles?

robwozniak: I know :smile: ok, imo adding props with custom class is quite ok for future development. Please let me know if you're ok with these changes

szafran89: let's assume it's ok for now ;)

PR #245, opened 2019-09-03
#53149 - Update banner component

jszczech:


src/atoms/banner/Banner.html
```diff
@@ -1,9 +1,5 @@
-<a
-  :href="link.href"
-  :aria-label="link.ariaLabel"
-  class="banner"
->
-  <picture class="image">
+<a class="a-banner">
```
szafran89: what if we want to replace a with `nuxt-link` or `route-link`?

src/atoms/banner/Banner.stories.js
```diff
@@ -1,51 +1,92 @@
...
+        default: text('Aria label', 'Banner')
+      },
+      hrefKnobs: {
+        default: text('Href', '#')
+      }
+    },
+    template: `
+      <a-banner
+        :href="hrefKnobs"
+        :ariaLabel="ariaLabelKnobs"
```
szafran89: `:aria-label`

PR #248, opened 2019-09-19
#57831 - Fix radio styles

jszczech:


szafran89:
Please also add knobs here

PR #283, opened 2020-07-03
Feature/update blockquote

ofrankowska:


src/atoms/blockquote/Blockquote.css
```diff
@@ -0,0 +1,7 @@
+blockquote,
```
szafran89: I think we can remove this line?

src/atoms/blockquote/Blockquote.vue
```diff
@@ -10,4 +10,4 @@ export default {
 }
 </script>
 
-<style lang="scss" src="./Blockquote.scss" />
+<style lang="css" src="./Blockquote.css" />
```
szafran89: remove lang attr

szafran89 (changes requested):

szafran89 (approved):

PR #284, opened 2020-07-19
Feature/update fieldset

ofrankowska:


src/atoms/fieldset/Fieldset.css
```diff
@@ -0,0 +1,11 @@
+.a-fieldset {
+}
```
szafran89: if we don't need set classes for block we can remove it.

szafran89 (changes requested):

szafran89 (approved):

szafran89 (approved):

PR #293, opened 2020-10-09
Feature/update breadcrumbs

ofrankowska:


src/atoms/breadcrumbs/Breadcrumbs.html
```diff
@@ -18,7 +25,11 @@
             :to="breadcrumb.href"
             :aria-label="breadcrumb.text"
```
anqaka: aria-label is not needed when we have a text inside the link :)

PR #312, opened 2021-05-21
Feature/update label

### LINKED ISSUE #264: Label component review
- review/update styles
- review/update UT
- review/update logic
- review/update a11y
- review/update stories and knobs

zuzannaszymanda:


src/atoms/label/Label.html
```diff
@@ -1,5 +1,5 @@
 <label
-  class="a-label"
+  :class="getClass('label')"
```
ofrankowska: when the element has only one attribute, then we don't split lines in general. can we stick to this convention?

src/atoms/label/Label.stories.js
```diff
@@ -1,30 +1,39 @@
...
-    textKnobs: {
-      default: text('Text', 'Label')
-    }
-  },
+  props: Object.keys(argTypes),
   template: `
-    <a-label :class="classKnobs">
-      {{ textKnobs }}
+    <a-label
+      :variant="variant"
```
ofrankowska: same as above

src/atoms/label/Label.stories.js
```diff
@@ -1,30 +1,39 @@
...
+        options: [
+          'primary',
+          'hidden'
+        ]
+      }
+    },
+    text: {
+      control: {
+        type: 'text'
+      }
```
ofrankowska: can you remove this code since the control works without any configuration?

src/atoms/label/Label.stories.js
```diff
@@ -1,30 +1,39 @@
...
+      :variant="variant"
+    >
+      {{ text }}
     </a-label>
   `
 })
+
+export const Default = Template.bind({})
+Default.args = {
+  text: 'Label'
```
ofrankowska: can you add the primary variant as the default one?

ofrankowska (changes requested):
can you update file paths using the new aliases?

ofrankowska (approved):

PR #315, opened 2021-06-04
Feature/306

### LINKED ISSUE #306: Input components review

szafran89:


src/molecules/input/Input.html
```diff
@@ -0,0 +1,34 @@
+<div :class="getClass('input')">
+  <!-- @slot Label (Named slot) -->
+  <slot name="label" v-bind="{ label, id }">
```
ofrankowska, line 3: formatting

src/molecules/input/Input.html
```diff
@@ -0,0 +1,34 @@
...
+    <a-label
+      :class="[
+        'transition-all ease-linear transition-300',
+        {
+          'text-xs inline-flex': variant ==='primary'
+        },
+        {
+          'flex-shrink-0 mr-4': variant !=='primary'
+        },
```
ofrankowska, line 12: why are these classes added here instead of config?

src/molecules/input/Input.html
```diff
@@ -0,0 +1,34 @@
...
+    v-on="listeners"
+    :value="value"
+    :id="id"
+    :class="[
+      getClass('input__field'),
+      value ? getClass('input__field__filled') : ''
+    ]"
+  >
+  <!-- @slot Icon (Named slot) -->
+  <slot name="icon" v-bind:variant="variant" />
```
ofrankowska, line 33: formatting

src/molecules/input/Input.spec.js
```diff
@@ -0,0 +1,92 @@
+import { mount, shallowMount } from '@vue/test-utils'
```
ofrankowska, line 1: `'shallowMount' is declared but its value is never read`

src/molecules/input/Input.stories.js
```diff
@@ -0,0 +1,130 @@
...
+    id: {
+      control: {
+        type: 'text'
+      }
+    },
+    label: {
+      control: {
+        type: 'text'
+      }
+    },
```
ofrankowska, line 19: text controls for props generate automatically so I think this can be removed

src/molecules/input/Input.stories.js
```diff
@@ -0,0 +1,130 @@
...
+        :id="id"
+        :variant="variant"
+      >
+        <template #icon="{ variant }">
+          <div
+            :class="[
+              'h-12 w-12',
+              'flex items-center justify-center',
+              'absolute right-0',
+              variant === 'primary' ? 'top-6' : 'top-0'
```
ofrankowska, line 83: I think it might be useful to have a variant in config for this. I can see also in designs that we have a seach input with icon on the left. Another slot and variant might be needed. 
image

ofrankowska (changes requested):
can you merge changes from `update`, update colors and add config in the new way?

PR #319, opened 2021-07-30
Feature/upgrade tailwindcss

anqaka:


src/atoms/container/Container.stories.js
```diff
@@ -1,39 +1,38 @@
...
-    <a-container :tag="tagKnob">
-      {{ contentKnob }}
-    </a-container>
-  `
-})
+export const Custom = Template.bind({})
+Custom.args = {
+  text: 'Container content',
+  tag: 'div'
+}
```
ofrankowska: with controls we can change the tag in the default story so I think this custom one isn't needed anymore

src/atoms/fieldset/Fieldset.html
```diff
@@ -1,13 +1,9 @@
...
-  class="a-fieldset"
   :aria-labelledby="`legend-${id}`"
 >
   <legend
     :id="`legend-${id}`"
-    :class="[
-      legendClass,
-      'a-fieldset__legend'
-    ]"
+    :class="getClass('fieldset-legend')"
```
ofrankowska: we use BEM in other components so I think we should stick to it

src/atoms/fieldset/Fieldset.stories.js
```diff
@@ -1,53 +1,32 @@
...
+  props: Object.keys(argTypes),
   template: `
     <a-fieldset
-      :legend-text="legendKnobs"
-      :legendClass="classKnob"
+      :legend-text="legendText"
+      :legend-hidden="legendHidden"
     >
-      <span>Fieldset content</span>
+      <div>Fieldset content</div>
```
ofrankowska: is this div needed?

src/atoms/fieldset/Fieldset.html
```diff
@@ -1,13 +1,9 @@
 <fieldset
```
ofrankowska, line 1: I think fieldset should have `:class="getClass` even if we don't add any classes on alpaca level so that they can be easily added in config without overriding the template

src/atoms/fieldset/Fieldset.stories.js
```diff
@@ -1,53 +1,32 @@
-import { select, text } from '@storybook/addon-knobs'
-
-import getClassKnobsConfig from '@utils/helpers/get-class-knobs-config.js'
-import selectorsConfig from './Fieldset.selectors.json'
-
 import AFieldset from './Fieldset.vue'
```
ofrankowska, line 1: for some reason I can't see fieldset in storybook 
image

src/atoms/loader/Loader.js
```diff
@@ -1,2 +1,45 @@
...
+        'm-auto',
+        'border-4', 'border-transparent', 'rounded-full',
+        'animate-spin'
+      ]
+    },
+    primary: {
+      loader__circle: [
+        'border-t-secondary'
+      ]
+    },
```
ofrankowska: shouldn't this be in base? it seems like loader doesn't look good without it

ofrankowska (changes requested):

ofrankowska (approved):

